PR #10, opened 2017-07-18
Options for mounting an artifacts dir and log file from the host

inercia:
Options for mounting an artifacts dir and log file from the host
Run tests in parallel


k8s-e2e-tests/entrypoint.sh
```diff
@@ -44,18 +44,31 @@ run_tests() {
...
+    env GINKGO_PARALLEL=y E2E_REPORT_PREFIX=parallel \
+    go run hack/e2e.go -v --test \
+        --test_args="$KUBE_TEST_ARGS --ginkgo.focus=\[Conformance\] --ginkgo.skip=\[Serial\]" 2>&1 | \
+            tee /tmp/e2e-test.log
+
+    # now run the serial tests
+    env E2E_REPORT_PREFIX=serial \
+    go run hack/e2e.go -v --test \
+        --test_args="$KUBE_TEST_ARGS --ginkgo.focus=\[Serial\].*\[Conformance\]" 2>&1 | \
+            tee -a /tmp/e2e-test.log
```
mjura: I am thinking loudly, should we split log files on Serial and Parallel? Anyway this change looks fine

inercia: As gar as I can remember, we must do that: otherwise they will overwrite...

PR #13, opened 2017-07-19
Fixes in the e2e tests

inercia:
New flag: `--no-tty`: do a `docker run` and `docker attach`
Split the `docker run` flags in several env vars.
New flag: `--no-build`: do not try to build the image
Do not try to modify the original `kubeconfig` passed to the `e2e-tests` runner
Do not fail if the certificates are not provided (they could be embeded in the `kubeconfig`)
Generate a unique container name (we could run several instances of this in Jenkins)
The `e2e-tests` runner can be run from a different directory.
Improved cleanups and error control

### DIFF
--- a/k8s-e2e-tests/e2e-tests
+++ b/k8s-e2e-tests/e2e-tests
@@ -1,8 +1,12 @@
 #!/bin/sh
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
 # options
 IMAGE_NAME=k8s-e2e-tests
-DOCKER_RUN_ARGS=${DOCKER_RUN_ARGS:-"-ti"}
+DOCKER_RUN_ARGS=${DOCKER_RUN_ARGS:-"--net=host"}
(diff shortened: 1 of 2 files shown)


kiall:
I'm not sure this is necessary. Infact, I think we just need to remove the `-t` flag from docker run, leaving `-i` in place. There is no need to allocate a TTY for this, as the e2e tests ask for no user input (-t is needed for things like bash, where you need to be able to type into the container... otherwise, it's not needed)

inercia:
> I'm not sure this is necessary. Infact, I think we just need to remove the -t flag from docker run, leaving -i in place. There is no need to allocate a TTY for this, as the e2e tests ask for no user input (-t is needed for things like bash, where you need to be able to type into the container... otherwise, it's not needed)

@kiall This PR only adds some configurability for running the tests. If you want to just use `-i`, you can. If you want to use some other flags, just do it. `docker run` and `docker attach` only will be used when you don't want any flags. For example, `-i` is not enough for running the tests with a ssh session. 

Besides that, this PR adds many other things that are necessary...

PR #18, opened 2017-07-20
drop workaround

MaximilianMeister:


MaximilianMeister:
code was not run anyways

PR #23, opened 2017-07-21
save screenshots

MaximilianMeister:


MaximilianMeister:
rebased

kiall:
What path do these get saved at? Ideally, we would have a screenshots folder which we can easily archive

MaximilianMeister:
> What path do these get saved at? Ideally, we would have a screenshots folder which we can easily archive

@kiall done

PR #29, opened 2017-07-26
save screenshot after error and timeout

MaximilianMeister:


MaximilianMeister:
an example run with a screenshot after failure can be found here https://github.com/kubic-project/automation/pull/29

other than that it seems that increasing the `wait` for the minions to be accepted has worked, although i still don't know why it takes so long (> 4min)

PR #30, opened 2017-07-26
Add terraform orchestration for CaaSP deployment on OpenStack cloud

mjura:
We would like to have CaaSP installation on OpenStack automated, terraform based can help with that.

### DIFF
--- a/caasp-openstack/README.md
+++ b/caasp-openstack/README.md
@@ -1,0 +1,37 @@
+# Orchestrate CaaSP on OpenStack
+
+This is terraform orchestration for CaaSP deployment on OpenStack cloud
+
+## Overview
+
+We would like to have CaaSP installation on OpenStack automated. This orchestration is helping with that. After providing data about OpenStack API's and credentials, CaaSP can be easely installed with one command. After CaaSP is installed it can be also scalled up by changing number of minion Workers.
+
+## Usage
+
+### Requirements
+
+Basic requirement is to have running OpenStack cloud environment. Next step is to get OpenAtack RC file with informations about OpenStack API url, user credentials and specific informations about DomaiName, RergionName and others.
+
+It is also required to create before deployment private network called, which name has to provided in options.
+
+#### Main options
+
+After downloading OpenStack RC file, please source openrc.sh file which is exporting OS variables with required informations. For more options please edit openstack.tfvars and change specific options about CaaSP deployment.
+
+#### Examples
+
+Deploy CaaSP on OpenStack
+```
+./caasp-openstack apply
+```
+
+Refresh current deployment and update existing cluster after changing workers number
+```
+./caasp-openstack plan
+./caasp-openstack apply
+```
+
+Cleanup installation
+```
+./caasp-openstack destroy
+```
--- a/caasp-openstack/caasp-cluster.tf
+++ b/caasp-openstack/caasp-cluster.tf
@@ -1,0 +1,265 @@
+variable "auth_url" {}
+variable "domain_name" {}
+variable "region_name" {}
+variable "project_name" {}
+variable "user_name" {}
+variable "password" {}
+variable "image_name" {}
+variable "network_name" {}
+variable "admin_size" {}
+variable "master_size" {}
+variable "masters" {}
+variable "worker_size" {}
+variable "workers" {}
+
+provider "openstack" {
+  domain_name = "${var.domain_name}"
+  tenant_name = "${var.project_name}"
+  user_name = "${var.user_name}"
+  password = "${var.password}"
+  auth_url = "${var.auth_url}"
+  insecure = "true"
+}
+
+data "template_file" "cloud-init" {
+  template = "${file("cloud-init.cls")}"
+
+  vars {
+    admin_address = "${openstack_compute_instance_v2.admin.access_ip_v4}"
+  }
+}
+
+resource "openstack_compute_keypair_v2" "keypair" {
+  name       = "caasp-ssh"
+  region     = "${var.region_name}"
+  public_key = "${file("ssh/id_caasp.pub")}"
+}
+
+resource "openstack_compute_secgroup_v2" "secgroup_base" {
+  name        = "caasp-base"
+  region      = "${var.region_name}"
+  description = "Basic security group for CaaSP"
+
+  rule {
+    from_port   = -1
+    to_port     = -1
+    ip_protocol = "icmp"
+    cidr        = "0.0.0.0/0"
+  }
+
+  rule {
+    from_port   = 22
+    to_port     = 22
+    ip_protocol = "tcp"
+    cidr        = "0.0.0.0/0"
+  }
+
+  rule {
+    from_port   = 2379
+    to_port     = 2379
+    ip_protocol = "tcp"
+    cidr        = "0.0.0.0/0"
+  }
+}
+
+resource "openstack_compute_secgroup_v2" "secgroup_admin" {
+  name        = "caasp-admin"
+  region      = "${var.region_name}"
+  description = "CaaSP security group for admin"
+
+  rule {
+    from_port   = 443
+    to_port     = 443
+    ip_protocol = "tcp"
+    cidr        = "0.0.0.0/0"
+  }
+
+  rule {
+    from_port   = 4505
+    to_port     = 4506
+    ip_protocol = "tcp"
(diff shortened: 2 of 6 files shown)


kiall:
This looks like it duplicates the stuff we already have in the openstack-heat folder? But, requires tools (terraform) is installed locally vs being able to use it from both the CLI and Horizon web UIs... 

Is there any reason we need to redo this with terraform instead of adding the extra pieces (security groups) to the heat templates?

MaximilianMeister:
> Is there any reason we need to redo this with terraform instead of adding the extra pieces (security groups) to the heat templates?

it's easier for the devs i think as we already have terraform installed with the devenv, e.g. i don't have any openstack clients installed on my machine. maybe some documentation on the usage of `openstack-heat` would help devs, but still i like more the terraform workflow, as we're using it all the time already (devenv, caasp-kvm)

plus we could maybe use it to replace the "terraform create environment" stage?

mjura:
Terreform is nice alternative for Heat. It can be used to orchestrate CaaSP on different Cloud providers in easy way (libvirt, OpenStack, AWS, GCE etc) it means that it is way for CaaSP deployment standardization. Other arguments, as Max said it is nicer for development. It can be used also to scaling CaaSP deployment, after changing number of Workers you have to call only `caasp-run plan` and `caasp-run apply` and new nodes will be added without redeploying whole CaaSP cluster. I will also submit changes for Heat templates.

mjura:
I would like to merge it, since I use it for my deployments

caasp-openstack/caasp-cluster.tf
```diff
@@ -0,0 +1,167 @@
...
+variable "user_name" {}
+variable "password" {}
+variable "image_name" {}
+variable "network_name" {}
+variable "admin_size" {}
+variable "master_size" {}
+variable "minions_size" {}
+variable "minions" {
+  default = 5
+}
```
kiall: Can we call these workers? both masters and workers are minions :)

caasp-openstack/caasp-cluster.tf
```diff
@@ -0,0 +1,167 @@
...
+  security_groups = [
+    "default",
+    "${openstack_compute_secgroup_v2.secgroup.name}",
+  ]
+
+  user_data = "${file("cloud-init.adm")}"
+}
+
+resource "openstack_compute_instance_v2" "master" {
+  name       = "CaaSP-Master"
```
kiall: We've been updating tooling to support multiple masters, can we add a count here to allow for multiple?

caasp-openstack/caasp-cluster.tf
```diff
@@ -0,0 +1,167 @@
...
+
+  flavor_name = "${var.admin_size}"
+  key_pair    = "caasp-ssh"
+
+  network {
+    name = "${var.network_name}"
+  }
+
+  security_groups = [
+    "default",
```
kiall: The default secgroup is almost universally "all everything" after people boot a few VMs, we should just rely on the groups we manage.

caasp-openstack/caasp-cluster.tf
```diff
@@ -0,0 +1,167 @@
...
+
+  flavor_name = "${var.master_size}"
+  key_pair    = "caasp-ssh"
+
+  network {
+    name = "${var.network_name}"
+  }
+
+  security_groups = [
+    "default",
```
kiall: The default secgroup is almost universally "all everything" after people boot a few VMs, we should just rely on the groups we manage.

caasp-openstack/caasp-cluster.tf
```diff
@@ -0,0 +1,167 @@
...
+
+  flavor_name = "${var.minions_size}"
+  key_pair    = "caasp-ssh"
+
+  network {
+    name = "${var.network_name}"
+  }
+
+  security_groups = [
+    "default",
```
kiall: The default secgroup is almost universally "all everything" after people boot a few VMs, we should just rely on the groups we manage.

caasp-openstack/caasp-cluster.tf
```diff
@@ -0,0 +1,167 @@
...
+  user_data = "${data.template_file.cloud-init.rendered}"
+}
+
+output "ip_dashboard" {
+  value = "${openstack_compute_instance_v2.admin.access_ip_v4}"
+}
+
+output "ip_master" {
+  value = "${openstack_compute_instance_v2.master.access_ip_v4}"
+}
```
MaximilianMeister: can we also add the admin ip here?

MaximilianMeister:
It would also be great if we can already associate floating ip's along with the nodes, and put that into the output of terraform

caasp-openstack/caasp-cluster.tf
mjura: I've changed the name of it on `workers`

caasp-openstack/caasp-cluster.tf
mjura: Ok, I've added three more security groups for each node role

caasp-openstack/caasp-cluster.tf
mjura: I've added also counter for master nodes and I'm waiting for this feature

caasp-openstack/caasp-cluster.tf
mjura: I've added output also for `ip_admin`

MaximilianMeister:
@mjura i think you forgot to check in the example config file `openstack.tfvars.example`

MaximilianMeister:
> i think you forgot to check in the example config file openstack.tfvars.example

oh i just saw the `.openrc` support, even better thanks

mjura:
I switched to `openrc` file with configuration, additional options can be changed in `caasp-cluster.tf` file. I've tested it on skycastle and works fine. I haven't added floating network yet. Probably in next PR's

caasp-openstack/caasp-cluster.tf
```diff
@@ -0,0 +1,275 @@
+variable "auth_url" {}
+variable "domain_name" {}
+variable "region_name" {}
+variable "project_name" {}
+variable "user_name" {}
+variable "password" {}
+variable "image_name" {
+  default = "CaaSP-1.0-Build30.25-GMC"
```
MaximilianMeister: i think we need some tfvars file after all, if only because of the image name

caasp-openstack/caasp-cluster.tf
MaximilianMeister: we should also show the `workers` i think, like in `caasp-kvm`. can you still add that?

MaximilianMeister:
MINOR: and @mjura could you rename `caasp-run` to `caasp-openstack` ? that would be the same schema like `caasp-kvm`

caasp-openstack/caasp-cluster.tf
mjura: I've added workers ip to output

mjura:
@MaximilianMeister: I've addressed all your comments. Floating IP's I will add later, I would like to have this initial change merge first. I've tested it and works fine. I use it all over the time for testing CaaSP

caasp-openstack/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+#### Main options
+
+After downloading OpenStack RC file, please source openrc.sh file which is exporting OS variables with required informations. For more options please edit openstack.tfvars and change specific options about CaaSP deployment.
+
+#### Examples
+
+Deploy CaaSP on OpenStack
+```
+./caasp-run apply
```
kiall: Minor: these should now be `./caasp-openstack ...`

caasp-openstack/openstack.tfvars
```diff
@@ -0,0 +1,7 @@
+image_name = "CaaSP-1.0-Build30.25-GMC"
+network_name = "caasp-net"
+admin_size = "m1.large"
+master_size = "m1.large"
+masters = 1
+worker_size = "m1.large"
+workers = 3
```
kiall: minor: setting this to 2 gives a valid, but smaller, cluster :)

mjura: Ok, change this default to 2

caasp-openstack/README.md
mjura: Fixed

mjura:
Ok, I've fixed minor comments. Thank you for reviewing

PR #32, opened 2017-07-26
Heat templates improvements for CaaSP

mjura:


openstack-heat/caasp-stack.yaml
```diff
@@ -39,7 +39,8 @@ parameters:
     type: string
     description: >
       ID of public network for which floating IP addresses will be allocated
-    default: 66ee2bc5-1464-4ce3-9386-52fe9a3d3518
+    constraints:
```
kiall: From memory, I had to remove this as the external net's aren't owned by your tenant, they don't show up.. I'll need to double check and will leave another comment later/tomorrow!

openstack-heat/caasp-stack.yaml
```diff
@@ -73,6 +74,45 @@ parameters:
...
+          port_range_max: 2379
+        - protocol: tcp
+          port_range_min: 2380
+          port_range_max: 2380
+        - protocol: tcp
+          port_range_min: 6443
+          port_range_max: 6443
+        - protocol: tcp
+          port_range_min: 10250
+          port_range_max: 10250
```
kiall: Should we also include Kubernetes nodeport range?  That's udp+tcp 30000-32768 by default.

openstack-heat/caasp-stack.yaml
grahamhayes: This should not be set, as due to both heat and neutron wierdness, we cannot select the external network from that constraints list

openstack-heat/caasp-stack.yaml
```diff
@@ -73,6 +74,45 @@ parameters:
...
+      rules:
+        - protocol: icmp
+        - protocol: tcp
+          port_range_min: 22
+          port_range_max: 22
+        - protocol: tcp
+          port_range_min: 443
+          port_range_max: 443
+
+  secgroup_cluster:
```
kiall: I'd like to split this group into master vs worker, as the ranges of required ports are different, we should try be a tight as possible.

openstack-heat/caasp-stack.yaml
```diff
@@ -107,6 +147,7 @@ resources:
       network_id: { get_param: int_net_id }
       security_groups:
         - default
```
kiall: We should remove the default group from here and the other roles, and leave just our own managed groups - that way, it's easier to notice when new ports are needed (as default almost always ends up with an "allow everything" rule)

openstack-heat/caasp-stack.yaml
grahamhayes: I tested it - the stack will not boot with the constraint - the ext net does not show up.

openstack-heat/caasp-stack.yaml
```diff
@@ -73,6 +74,45 @@ parameters:
...
+    type: OS::Neutron::SecurityGroup
+    properties:
+      rules:
+        - protocol: icmp
+        - protocol: tcp
+          port_range_min: 22
+          port_range_max: 22
+        - protocol: tcp
+          port_range_min: 443
+          port_range_max: 443
```
grahamhayes: We need to add the ectd ports here as well to allow the cluster to talk to the discovery etcd on the admin node

openstack-heat/caasp-stack.yaml
mjura: Ok, I agree, I will drop this commit

openstack-heat/caasp-stack.yaml
mjura: I added required range

openstack-heat/caasp-stack.yaml
mjura: I removed default security group

openstack-heat/caasp-stack.yaml
mjura: Done

mjura:
I've tested whole deployment with guestbook example and ingress loadbalancer on it. It works fine

mjura:
I've added two new commits, this PR was tested and works fine.

openstack-heat/caasp-stack.yaml
```diff
@@ -73,8 +77,104 @@ parameters:
     default: linux
 
 resources:
+  internal_network:
```
grahamhayes: On prvcld (the OpenStack cluster we are running in Provo) we have a limited amount of floating IPs. As a result, we want to only use the `shared-net` that is setup there.

(You loose an IP per neutron router created - and using the `shared-net` we limit this to a single IP across the cluster.)

kiall: Yea, we're a little limited on IPs in provo - but loosing 1 per stack is probably fine. We can always revisit later and spit it off/make it optional later.

openstack-heat/caasp-stack.yaml
```diff
@@ -35,16 +36,19 @@ parameters:
...
       - custom_constraint: glance.image
-  ext_net_id:
+  external_net:
     type: string
     description: >
-      ID of public network for which floating IP addresses will be allocated
-    default: 66ee2bc5-1464-4ce3-9386-52fe9a3d3518
-  int_net_id:
+      Name or ID of public network for which floating IP addresses will be allocated
+    default: floating
```
grahamhayes: this should be `ext-net`

kiall: Ah, I'm guessing the other internal clouds use floating by default.. I was just going to change ours to "floating" :)

grahamhayes: Cool - that works as well

PR #33, opened 2017-07-26
Reintroduce reset_sessions workaround

kiall:
This works around a possible issue with Puma handling multiple TLS
connections.

### DIFF
--- a/velum-bootstrap/spec/features/bootstrap_cluster.rb
+++ b/velum-bootstrap/spec/features/bootstrap_cluster.rb
@@ -13,6 +13,12 @@ feature "Boostrap cluster" do
     end
   end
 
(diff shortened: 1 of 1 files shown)


velum-bootstrap/spec/features/bootstrap_cluster.rb
```diff
@@ -13,6 +13,12 @@
     end
   end
 
+  # Using append after in place of after, as recommended by
+  # https://github.com/mattheworiordan/capybara-screenshot#common-problems
+  append_after(:each) do
```
MaximilianMeister: good catch, jfyi I silently dropped `capybara-screenshot` again as it was not really useful. but this might even help in our scenario

PR #47, opened 2017-08-03
Add tooling wrapper script

kiall:
Add a wrapper script to handle using all these tools quickly and easily.

We shouldn't put too much in here, if you need to customise the args passed to each tool, you should use the tool directly.

### DIFF
--- a/caasp-kvm/caasp-kvm
+++ b/caasp-kvm/caasp-kvm
@@ -1,0 +1,125 @@
+#!/bin/sh
+
+set -euo pipefail
+
+# options
+HAS_ACTION=false
+RUN_BUILD=false
+RUN_DESTROY=false
+
+MASTERS=1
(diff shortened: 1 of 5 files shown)


caasp-kvm/caasp-kvm
```diff
@@ -0,0 +1,125 @@
...
+  esac
+  shift
+done
+
+# Core methods
+build() {
+  log "CaaS Platform Building"
+
+  terraform apply -parallelism=$PARALLELISM \
+    -var "caasp_img_source_url=channel://${CHANNEL}" \
```
grahamhayes: This should allow us to use `file://` and `http(s)://`  as well - as `channel://` will always download the new file.

kiall: Done

PR #49, opened 2017-08-08
Tune number of tests that can run in parallel

mjura:
Compute number of Kubernetes ready cluster nodes and
adjust number of tests that can run in parallel depends on
what tests are running and on the size of the cluster.

Too many, and tests will fail due to resource contention.

### DIFF
--- a/k8s-e2e-tests/entrypoint.sh
+++ b/k8s-e2e-tests/entrypoint.sh
@@ -76,16 +76,28 @@ run_tests() {
         --clean-start=true \
         --dump-logs-on-failure \
         --delete-namespace=true \
-        --ginkgo.trace=true"
-#       --e2e-verify-service-account=false \
+        --ginkgo.trace=true \
+        --ginkgo.flakeAttempts=3"
+        # --ginkgo.flakeAttempts is the time to retry when there is a test failure.
 
     cd $KUBE_ROOT
 
+    # Compute number of ready Kubernetes cluster nodes
+    export K8S_NODES=`kubectl get nodes 2> /dev/null | grep " Ready " | wc -l`
+    export GINKGO_PARALLEL=y
+    # The number of tests that can be run in parallel depends on
+    # size of the Kubernetes cluster and number of worker nodes. 
+    # Too many, and tests will fail due to resource contention.
+    export GINKGO_PARALLEL_NODES=$K8S_NODES
+ 
     # first run the parallelizable tests
-    env GINKGO_PARALLEL=y E2E_REPORT_PREFIX=parallel \
+    env E2E_REPORT_PREFIX=parallel \
     go run hack/e2e.go -v --test \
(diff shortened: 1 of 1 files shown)


k8s-e2e-tests/entrypoint.sh
```diff
@@ -81,12 +81,23 @@ run_tests() {
...
     cd $KUBE_ROOT
 
+    # Compute number of ready Kubernetes cluster nodes
+    export K8S_NODES=`kubectl get nodes 2> /dev/null | grep " Ready " | wc -l`
+    export GINKGO_PARALLEL=y
+    export GINKGO_PARALLEL_NODES=$K8S_NODES
+    # The number of tests that can run in parallel depends on what tests
+    # are running and on the size of the cluster. 
+    # Too many, and tests will fail due to resource contention.
+    export GINKGO_PARALLELISM=$((7*$K8S_NODES))
```
kiall: My own curiosity, where did the number 7 come from?

kiall: Humm, I don't see any code in GINKGO to handle environment variables. Looking at https://github.com/vmware/kubernetes/blob/master/hack/ginkgo-e2e.sh - But most of them seem to (eventually) be passed as CLI arguments to Ginkgo itself (The exception I'm seeing is GINKGO_PARALLELISM).

I'm wondering if the original GINKGO_PARALLEL actually did anything at all?

mjura: It seems that this parameter doesn't have any influence on the tests. Instead of this we will use only `GINKGO_PARALLEL_NODES` which has real impact on number of parallel tests.

mjura: It seems that this parameter doesn't have any influence on the tests. Instead of this we will use only GINKGO_PARALLEL_NODES which has real impact on number of parallel tests.

mjura:
@kiall: Thank you for your comments, I've checked gingo test arguments and I have realized that number of parallel tests depend only on number of nodes, which can be configured by `GINKGO_PARALLEL_NODES` or `--nodes` option. 

I've added also `--ginkgo.flakeAttempts` option which is also time of retrying when test failed.

drpaneas:
I have tested these changes locally:

_Before these changes:_
Parallel Tests: FAIL! -- 116 Passed | 28 Failed | 0 Pending | 507 Skipped
Serial Tests: SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 648 Skipped PASS

_After these changes:_
Parallel Tests: SUCCESS! -- 114 Passed | 0 Failed | 0 Pending | 359 Skipped
Serial Tests: SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 470 Skipped PASS

/LGTM

kiall:
Taking Panagiotis's comment as the second +A and merging :)

mjura:
[(13) Live update from 1.5 to 1.7](https://trello.com/c/XaUwemy0/294-13-live-update-from-15-to-17)

PR #57, opened 2017-08-11
[1.0] Add terraform orchestration for CaaSP deployment on OpenStack cloud

MaximilianMeister:
backport of https://github.com/kubic-project/automation/pull/30


kiall:
I'm not sure this should be backported, it's probably harmless to, but will it be used?

MaximilianMeister:
> I'm not sure this should be backported, it's probably harmless to, but will it be used?

could be when someone wants to reproduce a bug for `1.0`, i also think it won't be used that often but it doesn't cost anything at least, so i'd say let's backport it

PR #59, opened 2017-08-14
KVM: Update admin node resources

kiall:
Set admin node to 3GB of RAM and 4 vCPU


kiall:
(caasp-kvm isn't tested in CI yet, so ignoring CI and merging anyway)

PR #60, opened 2017-08-14
velum-bootstrap: spec updated based on multiple master selection

vitoravelino:
I've just updated the selectors to reflect the change on velum for the current specs. No use case was added for the multiple master selection itself. Lemme if know if that's necessary or not.

Depends-On: https://github.com/kubic-project/velum/pull/278


kiall:
It's possible, but we haven't set it up yet :) For now, we can test locally..

Pulling the latest down now!

kiall:
Validated locally, works nicely!

PR #68, opened 2017-08-15
WIP: Downloader script cleanup

kiall:
* Move to misc-tools
* Download files to a download folder
* Put symlinks into that same download folder
* Disambiguate symlinks based on type  - we download both the velum container and the qcow2 from devel, this worked by fluke previously as we have the devel repo listed twice - once as devel, once as head.

### DIFF
--- a/caasp-kvm/cluster.tf
+++ b/caasp-kvm/cluster.tf
@@ -102,7 +102,7 @@ provider "libvirt" {
 
 resource "null_resource" "local_checkout_of_caasp_image" {
   provisioner "local-exec" {
-    command = "./tools/download_image.py ${var.caasp_img_source_url}"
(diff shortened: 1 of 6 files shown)


caasp-kvm/cluster.tf
```diff
@@ -102,7 +102,7 @@ provider "libvirt" {
 
 resource "null_resource" "local_checkout_of_caasp_image" {
   provisioner "local-exec" {
-    command = "./tools/download_image.py ${var.caasp_img_source_url}"
+    command = "./tools/download_image.py ../downloads kvm ${var.caasp_img_source_url}"
```
grahamhayes: Does this need to move to `../misc-tools/download_image.py` ?

kiall: Fixed

grahamhayes:
Done in #84

PR #70, opened 2017-08-15
Add support for floating network in OpenStack terraform

mjura:


mjura:
I've added also floating ips for Workers, after this change will be merged we can remove it once again https://github.com/kubic-project/salt/pull/188

PR #71, opened 2017-08-15
Wait for page to settle

kiall:
The "A supported deployment of .. " alert takes time to disapear, and
the button moves as a result. Capybara uses coordinates to trigger a
click, making it possible for us to "misclick" if the button moves


kiall:
Saving the screenshots for reference:

09_before_bootstrap_cluster


09_after_bootstrap_cluster

PR #77, opened 2017-08-16
Move local steps from TF -> caasp-kvm script

kiall:
This makes the output much more pleasant (e.g. wget output).

Additionally, fix download script to cleanup failed/partial downloads.

### DIFF
--- a/caasp-kvm/tools/download_image.py
+++ b/caasp-kvm/tools/download_image.py
@@ -88,16 +88,21 @@ def download_file(url, expected_name, force_redownload):
         # as corrupted
         remote_sha = requests.get(versioned_url + '.sha256')
 
-        os.system(
-            "wget %(url)s -O %(file)s --progress=dot:giga" %
-            { "url": versioned_url, "file": actual_name} 
-        )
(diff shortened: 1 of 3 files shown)


caasp-kvm/tools/download_image.py
```diff
@@ -88,16 +88,21 @@ def download_file(url, expected_name, force_redownload):
...
+            )
+            local_sha = os.popen('sha256sum %s' % actual_name).read()
+
+            if local_sha.split(' ')[0] not in remote_sha.text:
+                print("Local SHA: %s" % local_sha)
+                print("Remote SHA: %s" % remote_sha.text)
+                raise Exception("Download corrupted - please retry.")
+        except:
+            print("Deleting failed download")
+            os.unlink(actual_name)
```
grahamhayes: We haven't actually linked the files at this point, so the link will be pointing at a good file, so we should leave it

kiall: That's deleting the file which wget was downloading, rather than the symlink.

PR #80, opened 2017-08-16
Allow role tox envs to read environment.json

grahamhayes:


flavio:
@kiall can you figure out if this PR still makes sense?

PR #81, opened 2017-08-16
Split bootstrap_cluster.rb in two phases

FedericoCeratto:
Split sequence in two and add some waiting to prevent races.
Stop downcasing hostnames.

### DIFF
--- a/velum-bootstrap/spec/features/01-setup-velum.rb
+++ b/velum-bootstrap/spec/features/01-setup-velum.rb
@@ -1,0 +1,30 @@
+require "spec_helper"
+require 'yaml'
+
+feature "Register user and configure cluster" do
+
+  before(:each) do
+    unless self.inspect.include? "User registers"
+      login
+    end
+  end
+
+  # Using append after in place of after, as recommended by
+  # https://github.com/mattheworiordan/capybara-screenshot#common-problems
+  append_after(:each) do
+    Capybara.reset_sessions!
+  end
+
+  scenario "User registers" do
+    with_screenshot(name: :register) do
+      register
+    end
+  end
+
+  scenario "User configures the cluster" do
+    with_screenshot(name: :configure) do
+      configure
+    end
+  end
+
+end
--- a/velum-bootstrap/spec/features/02-bootstrap-cluster.rb
+++ b/velum-bootstrap/spec/features/02-bootstrap-cluster.rb
@@ -4,7 +4,7 @@ require 'yaml'
 feature "Boostrap cluster" do
 
   let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
-  let(:hostnames) { environment["minions"].map { |m| m["fqdn"].downcase if m["role"] != "admin" }.compact }
+  let(:hostnames) { environment["minions"].map { |m| m["fqdn"] if m["role"] != "admin" }.compact }
 
   before(:each) do
     unless self.inspect.include? "User registers"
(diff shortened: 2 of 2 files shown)


kiall:
We don't actually hardcode the usage of bootstrap cluster specifically, if  just name these `01-setup-velum.rb` and `02-setup-velum.rb` and remove the old one, it should pass.

(Though, another change just landed: https://github.com/kubic-project/automation/pull/60 )

kiall:
and .. https://github.com/kubic-project/automation/pull/82

FedericoCeratto:
Updated

velum-bootstrap/spec/features/02-setup-velum.rb
```diff
@@ -19,11 +19,7 @@
...
   end
 
-  scenario "User registers" do
-    with_screenshot(name: :register) do
-      register
-    end
-  end
+  # User creation has already been done in register_and_configure.rb
 
   scenario "User configures the cluster" do
```
kiall: This step was moved to `01-setup-velum.rb` so should be removed from here.. In theory, it's harmless, but, better to not do it twice :)

velum-bootstrap/spec/features/02-setup-velum.rb
```diff
@@ -34,6 +30,10 @@
   scenario "User accepts all minions" do
     visit "/setup/discovery"
 
+    puts ">>> Wait for minions to show up"
+    # FIXME: implement better detection
+    sleep 300
```
kiall: So - I suspect this is related to bare metal taking longer to provision and PXE boot? A few lines down is the check for all minions to be pending, if you add 300 seconds to it's wait time, it'll do the same thing, without a 5 min delay for faster environments.

velum-bootstrap/spec/features/02-setup-velum.rb
```diff
@@ -46,6 +46,11 @@
     end
     puts "<<< accept-all button enabled"
 
+    puts ">>> Wait to prevent query timeout"
```
kiall: I don't quite follow whats happening here - two sections prior to this waits for all the minions to show up, the section prior to this waits for the page to enable the accept all button (which can be a fraction of a second later, hence the wait), and the section after this just clicks it.

Between these steps, nothing else should be happening.. The page should be static/idle etc?

velum-bootstrap/spec/features/01-setup-velum.rb
```diff
@@ -0,0 +1,34 @@
+require "spec_helper"
+require 'yaml'
+
+feature "Register user and configure cluster" do
+
+  let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
+  let(:hostnames) { environment["minions"].map { |m| m["fqdn"] if m["role"] != "admin" }.compact }
+  let(:master_minion) { environment["minions"].detect { |m| m["role"] == "master" } }
```
kiall: Nit: None of these 3 get used in this spec, so we should remove them

kiall:
Another nit `02-setup-velum.rb` should be called `02-bootstrap-lcuster.rb` - velum was setup by `01-setup-velum.rb` :)

kiall:
Also needs https://github.com/kubic-project/automation/commit/0b1696f03b74134171ce49589d8af85bd05e3c36 sucked in

velum-bootstrap/spec/features/02-setup-velum.rb
FedericoCeratto: Comment added.

velum-bootstrap/spec/features/02-setup-velum.rb
FedericoCeratto: Updated

velum-bootstrap/spec/features/01-setup-velum.rb
FedericoCeratto: Removed

PR #84, opened 2017-08-17
Clean up downloader script

grahamhayes:
Reenables caching of KVM images
Only downloads velum images when the ETag changes
Downloads into <repo>/downloads
Moved the downloading tool to <repo>/misc-tools
Can download 2 current versions of caasp images


inercia:
You could also simplify the downloader by using `wget` with the right argument (see [this script](https://github.com/inercia/caasp-tf/blob/master/support/tf/download-image.sh))...

kiall:
Yea, the could be simplified more... but it's better than the current code :)

PR #86, opened 2017-08-18
Multi Master Fixes

kiall:
Three commits to fix booting multi master with caasp-kvm / velum-bootstrap.


robdaemon:
On first glance, it looks harmless, but I'll have to give it a run.

PR #89, opened 2017-08-22
KVM: Use real DNS names for Kube-API endpoint

kiall:
This allows for valid and working certificates when using multi-master
deployments.

### DIFF
--- a/caasp-kvm/tools/generate_environment.sh
+++ b/caasp-kvm/tools/generate_environment.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -eu
+
 chmod 600 tools/id_docker
+
+
+
 if command -v jq; then
     echo "Generating environment.json file"
     environment=$(cat terraform.tfstate | \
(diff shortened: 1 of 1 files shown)


kiall:
> ./tools/generate_environment.sh 3
	...
	> grep kubernetesHost environment.json 
	  "kubernetesHost": "kube-api-x3.devenv.caasp.suse.net",

	> ./tools/generate_environment.sh 2
	...
	> grep kubernetesHost environment.json 
	  "kubernetesHost": "kube-api-x2.devenv.caasp.suse.net",

	> dig +short kube-api-x3.devenv.caasp.suse.net
	10.17.2.1
	10.17.2.0
	10.17.2.2

	> dig +short kube-api-x2.devenv.caasp.suse.net
	10.17.2.0
	10.17.2.1

	> dig +short kube-api-x1.devenv.caasp.suse.net
	10.17.2.0

caasp-kvm/caasp-kvm
```diff
@@ -113,7 +113,7 @@ build() {
     -var "kubic_velum_dir=$(realpath ../../velum)" \
     -var "kubic_caasp_container_manifests_dir=$(realpath ../../caasp-container-manifests)"
 
-  ./tools/generate_environment.sh
+  ./tools/generate_environment.sh $MASTERS
```
grahamhayes: can we not count the masters instead of adding inputs to a file that has none currently?

grahamhayes:
One thing is that this now ties us to the R&D VPN to resolve the DNS entry. Are we OK with this?

kiall:
> One thing is that this now ties us to the R&D VPN to resolve the DNS entry. Are we OK with this?

We already do that for the images, so, I think it's OK until we really fix those across the board

PR #91, opened 2017-08-25
Option for disabling the kubeDNS deployment

inercia:
`kubeDNS` is already being deployed in the _CaaSP 2.0_ images, so we shouldn't need this...


kiall:
Should we not just remove this code, or at the very least disable it by default?

inercia:
> Should we not just remove this code, or at the very least disable it by default?

@kiall We don't know the capabilities of the kubernetes cluster we are testing against, and we are not forcing any particular version, so I think we should try to be flexible and accept any k8s cluster setup (things are different in Jenkins, but we don't have to restrict the possible usa cases for this code).

kiall:
I'm not sure I follow, our tools should be geared towards CaaSP, which has kube-dns on by default now - in Jenkins, and in every 2.0 customer deployment.

Supporting running this on any kubernetes cluster deployed using any mechanism isn't something we have the resources to handle, so we shouldn't pretend we can by making it an option (and then having to try fix it whenever it breaks)

inercia:
> I'm not sure I follow, our tools should be geared towards CaaSP, which has kube-dns on by default now - in Jenkins, and in every 2.0 customer deployment.

I mean this is perfectly capable of running the e2e tests against CaaSP 1.0 and I don't think it would consume any resources to keep that feature (it is just a flag).

kiall:
> I mean this is perfectly capable of running the e2e tests against CaaSP 1.0 and I don't think it would consume any resources to keep that feature (it is just a flag).

CaaSP 1.0 and 2.0 (and 3.0, 4.0) will diverge significantly, should we set the precedent now that tools in master should work against all CaaSP versions? Should every PR be validated against every version of CaaSP, to ensure it continues to function as new code is added etc?

Right now, some of the tools *happen* to work, but that's just because there was no divergence yet that's broken using `master` branch tools against `release-1.0` branch code.

This repo has a, and the k8s conformance test tool exists within, the `release-1.0` branch for use against 1.0 - `master` should remain focused exclusively on `master`, and `release-1.0` should remain focused exclusively on `release-1.0`.

inercia:
Ok, then I'll close the PR.

PR #94, opened 2017-08-28
Copy c-c-m to temp dir + mount into VM

grahamhayes:
This allows us to replace velums definitions
while allowing for the `__TAG__` replacement to happen in
the admin VM


caasp-kvm/injected-caasp-container-manifests/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
ereslibre: Is there a reason for not modifying the root `.gitignore` file? https://github.com/kubic-project/automation/blob/8849656358dfa8e7c52ca8865ff470a73bc01e61/.gitignore

grahamhayes: Not really - I would have needed to add an empty `.gitkeep` file there if I used the main git ignore file one, so I figured I could at least make the file useful :)

PR #96, opened 2017-08-29
Remove downloader etag code

kiall:
This is no longer necessary, as the docker image filenames now include
a build number. Remove this code to keep the codepath the same, and allow
the shasum to handle the verifications.


kiall:
Was +A'd then rebased to pull in a fix, merging.

PR #97, opened 2017-08-29
Adapt to use tags

jordimassaguerpla:


kiall:
Rebased over the other backports

PR #98, opened 2017-08-29
Backports 1.0

kiall:
Backports to 1.0.


kiall:
There are some multi-master commits here which are harmless to 1.0, but keeping the drift down makes future cherry picks easier

jordimassaguerpla:
I am not familiar with this code nor I can test it cause my caasp-kvm is "broken". However, if these are backports from master, I will trust you.

PR #101, opened 2017-08-31
Add specs to update CaaSP

kiall:
Add two new specs to handle upgrading of CaaSP, one for the admin node, and one
for the masters and workers.

Depends-On: https://github.com/kubic-project/jenkins-library/pull/76

### DIFF
--- a/velum-bootstrap/spec/features/03-upgrade-admin.rb
+++ b/velum-bootstrap/spec/features/03-upgrade-admin.rb
@@ -1,0 +1,66 @@
+require "spec_helper"
+require 'yaml'
+
+feature "Upgrade Admin Node" do
+
+  let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
+  let(:hostnames) { environment["minions"].map { |m| m["fqdn"] if m["role"] != "admin" }.compact }
+
+  before(:each) do
+    login
+  end
+
+  # Using append after in place of after, as recommended by
+  # https://github.com/mattheworiordan/capybara-screenshot#common-problems
+  append_after(:each) do
+    Capybara.reset_sessions!
+  end
+
+  scenario "User clicks update admin node" do
+    visit "/"
+
+    puts ">>> Wait for update-admin-node button to be enabled"
+    with_screenshot(name: :update_admin_node_button_enabled) do
+      expect(page).to have_selector("a.update-admin-btn", wait: 400)
+    end
+    puts "<<< update-admin-node button enabled"
+
+    puts ">>> Click to update admin node"
+    with_screenshot(name: :update_admin_node_button_click) do
+      find('a.update-admin-btn').click
+    end
+    puts ">>> update admin node clicked"
+
+    puts ">>> Wait for reboot-admin-node button to be enabled"
+    with_screenshot(name: :reboot_admin_node_button_enabled) do
+      expect(page).to have_selector("button.reboot-update-btn", wait: 15)
+    end
+    puts "<<< reboot-admin-node button enabled"
+
+    puts ">>> Click to reboot admin node"
+    with_screenshot(name: :reboot_admin_node_button_click) do
+      find('button.reboot-update-btn').click
+    end
+    puts ">>> reboot admin node clicked"
+
+    # Allow 30 seconds for Velum to be shutdown before proceeding
+    sleep 30
+
+    puts ">>> Wait for Velum to recover"
+    1.upto(600) do |n|
+      begin
+        visit "/"
+        expect(page.status_code).to be(200)
+        break
+      rescue Exception => e
+        if n == 600
+          raise e
+        else
+          puts "... still waiting for velum to recover"
+          sleep 1
+        end
+      end
+    end
+    puts "<<< Velum has recovered"
+  end
+end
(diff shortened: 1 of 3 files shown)


kiall:
Output:

	kiall@kiall-z620:~/SUSE/caasp/automation/velum-bootstrap> ./velum-interactions --update-admin --update-minions
	>>> [velum-bootstrap] Running Velum Interactions
	>>> User logs in
	<<< User logged in
	>>> Wait for update-admin-node button to be enabled
	<<< update-admin-node button enabled
	>>> Click to update admin node
	>>> update admin node clicked
	>>> Wait for reboot-admin-node button to be enabled
	<<< reboot-admin-node button enabled
	>>> Click to reboot admin node
	>>> reboot admin node clicked
	>>> Wait for Velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	... still waiting for velum to recover
	<<< Velum has recover
	.>>> User logs in
	<<< User logged in
	>>> Wait for update-all-nodes button to be enabled
	<<< update-all-nodes button enabled
	>>> Wait until all minions are pending an update
	<<< All minions are pending an update
	>>> Click to update all nodes
	>>> update all nodes clicked
	>>> Wait until update is complete
	<<< Update completed
	.

	Finished in 6 minutes 50 seconds (files took 0.59727 seconds to load)
	2 examples, 0 failures

	>>> [velum-bootstrap] Done

velum-bootstrap/spec/features/03-upgrade-admin.rb
```diff
@@ -0,0 +1,66 @@
...
+    puts ">>> reboot admin node clicked"
+
+    # Allow 30 seconds for Velum to be shutdown before proceeding
+    sleep 30
+
+    puts ">>> Wait for Velum to recover"
+    1.upto(600) do |n|
+      begin
+        visit "/"
+        expect(page.status_code).to be(200)
```
ereslibre: I'm not sure how this can work every time since if the expectation isn't met the first time (http status 503) the `visit` won't be retried and we will get an expectation error... @vitoravelino?

I think we would be safer to modify this logic with something like:

```
if page.status_code == 200
  break
end
```

so we are sure that we keep looping until we get a 200 or we retried too many times.

velum-bootstrap/spec/features/03-upgrade-admin.rb
```diff
@@ -0,0 +1,66 @@
+require "spec_helper"
+require 'yaml'
```
ereslibre: Let's use double quotes here too :)

velum-bootstrap/spec/features/04-upgrade-minions.rb
```diff
@@ -0,0 +1,51 @@
+require "spec_helper"
+require 'yaml'
```
ereslibre: Same here (quotes)

velum-bootstrap/velum-interactions
```diff
@@ -0,0 +1,135 @@
+#!/bin/sh
+
+set -euo pipefail
+
+# options
+HAS_ACTION=false
```
ereslibre: Minor (nit): Maybe initialize to `true` and on `-h` and `*)` set to `false`?

velum-bootstrap/spec/features/03-upgrade-admin.rb
kiall: This works because, when the expectation isn't met, an exception is raised - which we catch, and ignore, for up to 600 seconds (ish).

velum-bootstrap/velum-interactions
kiall: This is C+P from all the others, I'd like to keep them the same :)

velum-bootstrap/spec/features/04-upgrade-minions.rb
kiall: This is C+P from all the other specs, I'd like to keep them the same :)

velum-bootstrap/spec/features/03-upgrade-admin.rb
kiall: This is C+P from all the other specs, I'd like to keep them the same :)

kiall:
> About the explicit wait parameters I guess they are necessary for the tests to pass because sometimes we really need to wait more than others. Just wanted to make sure this is the case, so we don't end up with random wait times everywhere :)

The wait's are very much guesses, depending on the size of the cluster, different times are needed.

I plan to rework how wait's are calculated at some point, to be based on a min / max / node count multiplier - but later :)

(e.g. some per-node value multiplied by # of nodes, and bounded within the min / max)

ereslibre:
It's a +1 from me despite my comments. I agree that we should keep things the same and in any case tackle the issues I commented on all different tests (didn't know they were c&p).

PR #102, opened 2017-08-31
Add a velum-bootstrap wrapper script

kiall:
As new specs are added for upgrade, we want to ensure only the correct
specs are called at the correct time.


kiall:
re #101 - yea, I'll rebase it over this once this is in.

PR #118, opened 2017-09-05
KVM: Force VMs MTUs down to 1400

kiall:
In some envs, we're forced to drop the MTU down. Since we can't control this
with terraform's libvirt plugin, we can add it as an early run command with
cloud init. As this is a dev environment tool, a 1400 MTU should be harmless


kiall:
Failing CI due to RBAC changes partially landing, this should allow the last to merge - forcing though,

PR #119, opened 2017-09-06
Add HTTP proxy support for dev tools

kiall:
Setting CAASP_HTTP_PROXY in ~/.bashrc or similar should mean the dev tools will use the proxy to download anything it needs to fetch from NUE etc

Depends-On: https://github.com/kubic-project/jenkins-library/pull/82

### DIFF
--- a/caasp-devenv
+++ b/caasp-devenv
@@ -17,6 +17,7 @@ RUN_DESTROY=false
 
 MASTERS=1
 WORKERS=2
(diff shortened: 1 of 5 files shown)


kiall:
Please ensure https://github.com/kubic-project/jenkins-library/pull/82 is merged BEFORE this change, otherwise CI will start pulling directly from NUE instead of via the proxy.

PR #123, opened 2017-09-07
Heat: Small tracking fix, update support

kiall:
Heat: Small tracking fix, update support

### DIFF
--- a/caasp-openstack-heat/.gitignore
+++ b/caasp-openstack-heat/.gitignore
@@ -1,3 +1,3 @@
-.stack_id
+.stack_name
 environment.json
 heat-environment.yaml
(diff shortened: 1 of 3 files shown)


kiall:
> I've tested this change and works fine, but I have doubts about splitting Heat files, because it breaks Horizon Orchestration way of deploying CaaSP.

I couldn't find another way to allow for arbitrary master numbers, where those masters are also getting floating IPs. I'll take another look on Monday.

kiall:
I'm going to split off the first two commits, as I'd like to see those land... The masters one is a little harder, I can't find a obvious way to do what we need, while keeping Horizon working.

kiall:
Multi master split off into https://github.com/kubic-project/automation/pull/136

kiall:
@mjura / @FedericoCeratto - could you guys have another look? I've stripped out the multi-master piece as there were some concerns there, the other two commits should be good to go though.

PR #124, opened 2017-09-08
Add bare metal deploy script

FedericoCeratto:
Add bare metal deployment together with https://github.com/kubic-project/jenkins-library/pull/80

### DIFF
--- a/caasp-bare-metal/deployer/deploy_testbed.py
+++ b/caasp-bare-metal/deployer/deploy_testbed.py
@@ -1,0 +1,671 @@
+#!/usr/bin/env python3 -Wd -b
+
+from argparse import ArgumentParser
+import base64
+from datetime import timedelta, datetime
+from time import sleep, time
+import http.client
+import json
+import logging
+import os.path
+import socket
+import subprocess
+import sys
+import urllib.parse
+
+import yaml
+from prometheus_client import CollectorRegistry, Summary, push_to_gateway
+from prometheus_client import Counter
+
+from environment_json import create_environment_json
+# import velum_client
+
+conf = None
+
+class TimedFormatter(logging.Formatter):
+    def format(self, record):
+        t = timedelta(seconds=int(record.relativeCreated / 1000))
+        record.elapsed = str(t)
+        return super(TimedFormatter, self).format(record)
+
+
+HWManager = None  # iLO or IPMI proxied by Bare Metal Manager
+
+log = logging.getLogger()
+log.setLevel(logging.DEBUG)
+handler = logging.StreamHandler()
+handler.setFormatter(TimedFormatter('%(elapsed)s %(message)s'))
+log.addHandler(handler)
+
+# This is used to grab servers from previous deployments that never
+# released their hardware and is used only when there are not enough spare
+# servers available.
+SERVERS_TIMEOUT = 7200
+
+AUTOYAST_URL_PATH = "/autoyast"
+AUTOYAST_SIG_CHUNK = """\
+    <signature-handling>
+      <accept_file_without_checksum config:type="boolean">true</accept_file_without_checksum>
+      <accept_non_trusted_gpg_key config:type="boolean">true</accept_non_trusted_gpg_key>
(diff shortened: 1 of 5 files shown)


caasp-bare-metal/deployer/environment_json.py
```diff
@@ -0,0 +1,109 @@
...
+           "proxyCommand" : "ssh root@10.84.72.172 -W %h:%p",
+           "addresses" : {
+              "privateIpv4" : "172.24.0.8"
+           },
+           "fqdn" : "test2-worker-eorkpgetfvr6-4rzhx3nww5fv-glz5hrejhktm.k8s.local",
+           "role" : "worker",
+           "index" : "3"
+        }
+     ]
+  }
```
kiall: These examples get out of date pretty fast :) Can we update the comment to point at the sample in the root of this repo instead? That way we just "maintain" one example.

caasp-bare-metal/deployer/deploy_testbed.py
```diff
@@ -0,0 +1,753 @@
...
+        cmd = cmd.split()
+    subprocess.check_call(cmd)
+
+def runcmd_out(cmd):
+    if isinstance(cmd, str):
+        cmd = cmd.split()
+    return subprocess.check_output(cmd).splitlines()
+
+def create_pods(podname, admin_host_ipaddr, master_ipaddr):
+    ## Create Kubernetes pods
```
kiall: Since we want these kind of tests to be generic, and suitable for use on all env types, can we remove these for now?

caasp-bare-metal/deployer/deploy_testbed.py
```diff
@@ -0,0 +1,753 @@
...
+
+    tsclient.list_pods(master_ipaddr)
+
+    log.info("Creating pods")
+    tsclient.create_pods(master_ipaddr, podname)
+
+    tsclient.list_pods(master_ipaddr)
+
+def scale_up_pods(podname, pods_num, master_ipaddr):
+    """Scale up/down pods replicas"""
```
kiall: Since we want these kind of things to be generic, and suitable for use on all env types, can we remove these for now?

caasp-bare-metal/deployer/deploy_testbed.py
```diff
@@ -0,0 +1,753 @@
...
+                use_bogus_hosts=True)
+
+    if args.velum_setup:
+        # Required in order to fetch /autoyast later on
+        log.info("Configuring Velum at http[s]://{} ".format(
+            admin_host_ipaddr))
+        generate_environment_json(admin_host_ipaddr, [],
+                use_bogus_hosts=True)
+        run_velum_client("01-setup-velum.rb")
+        log.info("Velum setup - done")
```
kiall: Ideally, the different tools wouldn't get "mixed" together, it adds to the maintenance. (e.g. when 01- is renamed, as new steps are added, etc).. Can we strip out the unused (in CI) stuff for the moment?

I know there also the QA use case to handle, but I think that would be better suited to a simple + separate to the bare-metal tool script that calls the commands from each tool in the right order - we do this with the `caasp-devenv` script in the root of this repo already for KVM environments.

caasp-bare-metal/deployer/deploy_testbed.py
```diff
@@ -0,0 +1,753 @@
...
+        admin_node = load_serverlist(args.testname, want_admin=True,
+            want_nodes=False)[0]
+        servername, serial, desc, ilo_ipaddr, ilo_iface_macaddr, eth0_macaddr = admin_node
+        power_up_time = datetime.now() - timedelta(hours=4)
+        admin_host_ipaddr = parse_dhcp_logs(power_up_time, eth0_macaddr)
+
+    assert admin_host_ipaddr
+
+    if args.bogus_env_json:
+        generate_environment_json(admin_host_ipaddr, [],
```
kiall: Can we just call this by default as the admin node is deployed? It's cheap to run, any the rest of the tools generate these right as the create the machines.

PR #128, opened 2017-09-14
Add pod scaling scripts and confs

FedericoCeratto:
Add a simple test pod.


mjura:
Can you please change directory to for example /k8s-pod-tests and update README about how to test it and what results we should expect.

Thank you

flavio:
pinging @FedericoCeratto to move ahead with this PR

PR #129, opened 2017-09-15
Remove credentials error from autoyast file

FedericoCeratto:
Related to https://github.com/kubic-project/velum/pull/270


ereslibre:
A bug report has been created regarding this issue: https://bugzilla.suse.com/show_bug.cgi?id=1059016

I opened a PR that I still need to validate: https://github.com/kubic-project/velum/pull/309

I think if we take into account this timeout and just let the warning be automatically dismissed we wouldn't need to remove this text (aside that the text itself will change with the new timeout on the XML, and this substitution won't help).

PR #131, opened 2017-09-18
Velum Bootstrap: Vary timeouts based on node count

kiall:
Vary the timeouts based on the # of nodes to allow for larger clusters to be deployed
without unnecessarily hitting the timeout.

The second commit removes another unnecessary check which is inherently prone to race conditions.

### DIFF
--- a/velum-bootstrap/spec/features/02-bootstrap-cluster.rb
+++ b/velum-bootstrap/spec/features/02-bootstrap-cluster.rb
@@ -52,9 +52,11 @@ feature "Boostrap cluster" do
     sleep 30
     visit "/setup/discovery"
 
-    puts ">>> Wait until Minion keys are accepted by salt"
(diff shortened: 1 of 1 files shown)


kiall:
> some minor enhancement suggestion: as they are static values throughout the test, could you add them as global variables via let(:accept_timeout) { ... } and orchestration_timeout?

Given there only used once, it made sense to calculate them next to their usage rather than add global variables?

MaximilianMeister:
> Given there only used once, it made sense to calculate them next to their usage rather than add global variables?

it's just a minor thing, feel free to ignore. generally it's just good practice to use `let` to make use of rspec's internal caching and lazy loading, and you don't have to look for the variable definition in the test when you want to adapt it

kiall:
> generally it's just good practice to use let to make use of rspec's internal caching and lazy loading

That makes sense for things that get used over+over, but IMO, not for something like this, keeping the code flow readable is IMO better in this case :)

PR #141, opened 2017-09-21
Heat: Regenerate environment JSON when updating

kiall:


caasp-openstack-heat/caasp-openstack
```diff
@@ -132,6 +132,9 @@ update_stack() {
   openstack stack update --wait -e $HEAT_ENVIRONMENT_FILE -t caasp-stack.yaml $stack_name \
     --parameter worker_count=$WORKERS \
     --parameter image=$IMAGE
+
+  ./tools/generate-environment $NAME
+  ../misc-tools/generate-ssh-config environment.json
```
inercia: It seems to me that these two scripts are always used in conjunction, am I right?. Maybe we could merge them...

kiall: Each tool (e.g. kvm v openstack v bare metal) has a different implementation of generate-environment, and the ssh-config is then generated from that environment file - so we can't merge them (though, we could call the SSH config gen from the various generate-environment.sh's

inercia: I think we could do the same thing [kubernetes does for cluster providers](https://github.com/kubernetes/kubernetes/tree/master/cluster) with `KUBERNETES_PROVIDER`: we could define, for example, `CAASP_PROVIDER=caasp-kvm` and then create a global `generate-environment` that does `${CAASP_PROVIDER}/generate-environment && generate-ssh-config`...

PR #144, opened 2017-09-25
WIP: Improve development environment with `caasp-kvm`

ereslibre:
The `caasp-kvm` does not gracefully handle a complete development environment:

* Generate migrations without having rails installed in my computer, edit them locally, and migrate them.
* When running migrations/rollbacks, update my `schema.rb` file on the velum project in my host.
* When running `rspec`, update `coverage` folder on the velum project in my host. This includes very detailed information about the overall code coverage.
* SSH into the admin node with a single, short command, so we can inspect it.

When we have this very "rare" two-way modifications things aren't very smooth, but the `caasp-devenv` script in this repository might help to make things easier.

This is just a way to fix all the workflows that I found broken with `caasp-kvm`. This is in some way an evolving PR and I'd like to know if you think this is the right place for this kind of changes or we should do something different.

- [ ] How to gracefully modify the `Gemfile.lock` and work with that new versions of the gems or new gems?


robdaemon:
This seems really, really useful. My only request is the ability to pass arbitrary arguments to RSpec and rubocop. When I'm working on an individual feature, I tend to run the tests for only that feature.

MaximilianMeister:
I also really like the idea, but wouldn't a simple `--exec-on-admin` be sufficient? at least it's easier to maintain IMO

ereslibre:
Closing this for now.

PR #146, opened 2017-09-26
Remove some hardcoded paths

inercia:
Add some aliases for the ssh config file
Copy some stuff to the admin node (like scripts)
Some code cleanups


caasp-devenv
```diff
@@ -1,33 +1,46 @@
...
+MASTERS=${CAASP_NUM_MASTERS:-1}
+WORKERS=${CAASP_NUM_WORKERS:-2}
+IMAGE=${CAASP_IMAGE:-channel://devel}
+PROXY=${CAASP_HTTP_PROXY:-}
+PARALLELISM=${CAASP_PARALLELISM:-1}
+
+# base distribution used for setting up dependencies, and the packages to install
+DIST_BASE="opensuse.org/repositories/Virtualization:/containers"
+DIST_PACKAGES=jq \
+              python-tox \
```
MaximilianMeister: it fails here with
```
./caasp-devenv: line 35: python-tox: command not found
```

caasp-devenv
```diff
@@ -77,18 +90,16 @@ USAGE
...
 error()      { log "ERROR: $@" ; exit 1 ; }
-check_file() { if [ ! -f $1 ]; then error "File $1 doesn't exist!"; fi }
+check_file() { [ -f "$1" ] || error "File $1 doesn't exist!" ; }
 
 # parse options
 while [[ $# > 0 ]] ; do
   case $1 in
     --setup)
       RUN_SETUP=true
-      HAS_ACTION=true
```
kiall: The HAS_ACTION stuff was there in order to print out the help when the script hasn't been told to do anything, any reason we're removing that?

caasp-devenv
```diff
@@ -100,19 +111,17 @@ while [[ $# > 0 ]] ; do
...
     -d|--destroy)
       RUN_DESTROY=true
-      HAS_ACTION=true
       ;;
     -i|--image)
       IMAGE="$2"
       shift
       ;;
     -B|--bootstrap)
+      RUN_BUILD=true
```
kiall: Bootstrap should only bootstrap (and, same for all the commands really - 1 flag == 1 action, if you want 2 actions, specify 2 flags). I know I use the bootstrap flag several times in a row without rebuilding (certain failures will allow you to retry the bootstrap / velum-interactions phases).

caasp-devenv
```diff
@@ -130,81 +139,58 @@ while [[ $# > 0 ]] ; do
...
-fi
-
-if [ "$RUN_TESTINFRA" = true ] ; then
-  testinfra
-fi
-
-if [ "$RUN_DESTROY" = true ] ; then
-  destroy
-fi
+[ -n "$RUN_SETUP"     ] && setup
```
kiall: I'm not sure I understand the value of switching from simple true / false values and tests to testing if the variables are non-null? The code as it was before was trivial to glance at and understand exactly what to expect?

caasp-kvm/caasp-kvm
```diff
@@ -2,35 +2,47 @@
...
-    -p|--parallelism                 Set terraform parallelism
-    -P|--proxy                       Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+    -p|--parallelism             Set terraform parallelism
+    -P|--proxy                   Set HTTP proxy (Default: CAASP_HTTP_PROXY)
+
+  * Directories
+
+     --salt-dir       <DIR>      Local checkout of the Salt directory
+     --manifests-dir  <DIR>      Local checkout of the manifests directory
+     --velum-dir      <DIR>      Local checkout of the Velum directory
```
kiall: Can we mention the env variables here (similar to what we did for HTTP proxy?) - Ideally, anything you can set which influences the script should be detailed in the help output, i.e. we're trying to avoid "hidden" variables that you have to read the code or be told about to know they exist.

caasp-kvm/caasp-kvm
```diff
@@ -93,64 +115,73 @@ while [[ $# > 0 ]] ; do
...
+    rm -rf $injected/*
+    cp -r $CAASP_MANIFESTS_DIR/* $injected/
 
-  log "Copying CaaSP Container Manifests"
-  rm -rf $(realpath injected-caasp-container-manifests)/*
-  cp -r $(realpath ../../caasp-container-manifests)/* $(realpath injected-caasp-container-manifests)/
+    log "Patching Container Manifests"
+    $DIR/tools/fix-kubelet-manifest -o $injected/public.yaml $injected/public.yaml
+  else
+    log "Skipping Velum environment"
```
kiall: Oh, nice - does it actually work when the velum parts are skipped? Making those optional was something I wanted to do, but hadn't got around to... I figured it would be harder than it looked!

caasp-kvm/caasp-kvm
```diff
@@ -93,64 +115,73 @@ while [[ $# > 0 ]] ; do
...
-  terraform destroy -force -parallelism=$PARALLELISM \
-    -var "kubic_salt_dir=$(realpath ../../salt)" \
-    -var "kubic_velum_dir=$(realpath ../../velum)" \
-    -var "kubic_caasp_container_manifests_dir=$(realpath ../../caasp-container-manifests)"
+  log "Destroying terraform configuration"
+  terraform destroy -force $TF_ARGS
 }
 
-# main
-if [ "$HAS_ACTION" != true ]; then
```
kiall: Same Q as above - why are removing the help output when you don't specify an action?

caasp-devenv
inercia: Fixed. Thanks.

caasp-devenv
inercia: Ok, I thought we could not bootstrap without building first. I will revert it.

caasp-devenv
inercia: I think it is more _bash_-like to check for vars instead of equality with some predefined constant. `true` can always lead to confusion (_"was it True? or maybe TRUE? or maybe 1?"_, etc) as shell scripts do not have the idea of booleans. So it is usually safer to check `var=something`/`var empty` instead of `var=some-constant`.

caasp-kvm/caasp-kvm
inercia: Yeah, I will do that, but these env vars are intended to be used from scripts that call this script (ie, `caasp-devenv`) where they will be hidden for users anyway...

caasp-kvm/caasp-kvm
inercia: Ok, changed.

PR #149, opened 2017-09-29
caasp-kvm: Be more explicit about requirements

bear454:
Setup instructions should assume a clean install, not a well-worndeveloper's environment, in order to cover all necessary changes.

These points were derived from a series of failed attempts to install caasp-kvm on a newly installed Leap system, eventually leading to success.


kiall:
Thanks @bear454!

PR #151, opened 2017-10-03
Remove unused terraform variable

flavio:
The variable specifying the manifests directory is no longer used.

### DIFF
--- a/caasp-kvm/caasp-kvm
+++ b/caasp-kvm/caasp-kvm
@@ -142,7 +142,6 @@ fi
 
 if [ -n "$CAASP_MANIFESTS_DIR" ] ; then
   CAASP_MANIFESTS_DIR="$(realpath $CAASP_MANIFESTS_DIR)"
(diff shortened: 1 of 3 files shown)


kiall:
How do the updated manifests end up injected if this is no longer used?

caasp-kvm/caasp-kvm
```diff
@@ -140,12 +140,6 @@ if [ -n "$CAASP_VELUM_DIR" ] ; then
   log "Using Velum dir: $CAASP_VELUM_DIR"
 fi
 
-if [ -n "$CAASP_MANIFESTS_DIR" ] ; then
-  CAASP_MANIFESTS_DIR="$(realpath $CAASP_MANIFESTS_DIR)"
-  TF_ARGS="$TF_ARGS -var kubic_caasp_container_manifests_dir=$CAASP_MANIFESTS_DIR"
```
kiall: Actually, we should leave this block - but remove the TF_ARGS. It's still good to log which path we're using etc

flavio: Good point, I left the log line

flavio:
Changes done

PR #152, opened 2017-10-04
Fix testinfra requirements (cffi 1.11.1 is failing on us)

ereslibre:


kiall:
Upstream cfffi bug that's that likely trigger: https://bitbucket.org/cffi/cffi/issues/339/no-wheel-packages-for-1111-version

PR #153, opened 2017-10-04
[WIP] Reorganize the cluster providers

inercia:
The `caasp-devenv` in the `<root>` can invoke `provider/$CAASP_PROVIDER/caasp-{up|down}.sh` now (in the same fashion as the [kubernetes cluster providers](https://github.com/kubernetes/kubernetes/tree/master/cluster)). This will give us a way to have generic methods that can be overwritten for specific providers...


kiall:
In general, I like it (and planned to do the same myself at some point..) but it's pretty disruptive (every other PR will need rework, it'll need jenkins changes too etc). Can we wait till 2.0 is out and then pick a date to have the breakage? :)

I'd also prefer each tool remains having it "main script" name as they are already, up/down.sh can proxy to the right thing?

At least for me, I use those scripts more often than I use the top level caasp-devenv script!

inercia:
> In general, I like it (and planned to do the same myself at some point..) but it's pretty disruptive (every other PR will need rework, it'll need jenkins changes too etc). Can we wait till 2.0 is out and then pick a date to have the breakage? :)

Yeah, I'm not sure about this whole PR. As you said, it is very disruptive. My original intention was to move the `environment.json` to a common directory (so it could be reused by the e2s tests regardless of the cloud provider used) and to have `caasp-up.sh` and `caasp-down.sh` scrips that could be used instead of the `kvm` equivalents.

> I'd also prefer each tool remains having it "main script" name as they are already, up/down.sh can proxy to the right thing? At least for me, I use those scripts more often than I use the top level caasp-devenv script!

I think you are always using `caasp-kvm` because of the defects of the whole solution. If the _proxy_ is written in the right way, you could always use it instead of the `caasp-<provider>`s. For example, it could just take the arguments it wants and pass the rest to the proxied scripts...

inercia:
Closing because of too many conflicts.

PR #157, opened 2017-10-06
caasp-openstack: enable passing name to subcommands

MaximilianMeister:
when you have deployed multiple stacks this is helpful as the
.stack_name file will be outdated


MaximilianMeister:
i also thought about creating `.stack_name-$name` but this is a quicker fix

MaximilianMeister:
> Why do we need rm with force flag?

it's not that important, but if the file doesnt exist the script will fail.

PR #166, opened 2017-10-10
KVM: Allow RAM and CPU per node type to be changed

kiall:


caasp-kvm/caasp-kvm
```diff
@@ -19,6 +19,13 @@ CAASP_SALT_DIR=${CAASP_SALT_DIR:-$DIR/../../salt}
 CAASP_MANIFESTS_DIR=${CAASP_MANIFESTS_DIR:-$DIR/../../caasp-container-manifests}
 CAASP_VELUM_DIR=${CAASP_VELUM_DIR:-$DIR/../../velum}
 
+ADMIN_RAM=${CAASP_ADMIN_RAM:-4096}
+ADMIN_CPU=${CAASP_ADMIN_CPU:-4}
+MASTER_RAM=${CAASP_MASTER_RAM:-2048}
+MASTER_CPU=${CAASP_MASTER_CPU:-2}
+WORKER_RAM=${CAASP_WORKER_RAM:-2048}
+WORKER_CPU=${CAASP_WORKER_CPU:-2}
```
kiall: These are the defaults in `cluster.tf`. I suspect the admin node is too large... Lets merge this, let people find values that work better for them, and then update both here and cluster.tf with "better" defaults.

PR #170, opened 2017-10-13
Velum-bootstrap should use the correct URL to access Velum

kiall:
Velum bootstrap should be using the same name we expect end-users of
CaaSP to use when accessing the cluster.


MaximilianMeister:
feel free to merge, so we could gate https://github.com/kubic-project/velum/pull/331 as the first candidate

PR #173, opened 2017-10-18
Make remote libvirt hosts scenario easier

flavio:
## Why

I do all the development on my laptop but I cannot run a caasp cluster because on it because it's not powerful enough.

However I do have a powerful workstation that I can use. By default the devenv creates a cluster located inside of a natted network that is reachable only by the virtualization host.

I could workaround that by using a ssh tunnel, but this is suboptimal, especially when I want to play with applications running on top of the kubernetes cluster.

## The changes

With this change the VMs are still created inside of a dedicated network managed by libvirt, but they are also reachable by the outside (once a dedicated routing rule is added to the host).

For example, given the hypervisor host ip address is `192.168.1.39` and all the machines are on the `192.168.200.0/24` network, the following command will allow another host to access these VMs:

```
ip -4 route add 192.168.200.0/24 via 192.168.1.39
```

Changing the network type from `nat` to `route` is not enough: the default subnet is used also by SUSE's R&D network. My laptop is always connected to this network, hence all the requests to the
`10.17.0.0/22` network are automatically routed to the VPN instead of my workstation.

This commit allows also to change the default subnet to use a different one that is not conflicting with SUSE's R&D network (like the `172.30.0.0/22` one).

**WARNING:** you cannot pick networks like `172.16.0.0/13` or `172.17.0.0/24` because the first one is being used by flannel, while the second one is used by docker. You should use something like `172.30.0.0/22`.

## My dev enviroment/workflow

For the sake of completeness (and reuse from others), all the source code is checked out on my laptop inside of `~/hacking/kubic`. This is where I make all my changes.

The whole `~/hacking/kubic` directory is automatically sent to my workstation (hostname `arrakis`) and kept in sync using [lsyncd](https://github.com/axkibe/lsyncd) (which has official openSUSE packages BTW).

The configuration file for `lsyncd` is saved inside of `~/hacking/kubic/lsyncd.conf`:

```lua
settings {
  logfile    = "/tmp/lsyncd.log",
  statusFile = "/tmp/lsyncd.status",
  nodaemon   = true,
}

sync {
  default.rsyncssh,
  source    = "/home/flavio/hacking/kubic",
  host      = "flavio@arrakis",
  targetdir = "/home/flavio/hacking/kubic",
  exclude   = {
    "automation/downloads/**",
    "*.tfstate",
    "*.tfstate.backup",
    "*.terraform" }
}
```

## Bonus points

In addition to making the cluster easier to be accessed, it's now possible to create bigger cluster spanning several virtualization hosts.

For example: I can bring up the admin node, 3 masters and 1 worker on my workstation then add 3 workers nodes hosted on my laptop. 😎

### DIFF
--- a/caasp-kvm/README.md
+++ b/caasp-kvm/README.md
@@ -110,7 +110,7 @@ Clone these repositories:
 
       ./caasp-kvm --destroy
 
-## Using a cluster
+## Using a cluster from the hypervisor node
 
 The nodes booted will be given fixed/deterministic IP addresses, it it recommended
 you create a ~/.ssh/config with the following content, adjusting the SSH key path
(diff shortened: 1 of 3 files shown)


flavio:
Updated the PR description, everything is working fine. It was just me using a wrong value for the subnet.

caasp-kvm/cluster.tf
```diff
@@ -320,7 +326,7 @@ resource "libvirt_domain" "worker" {
   network_interface {
     network_id     = "${libvirt_network.network.id}"
     hostname       = "caasp-worker-${count.index}"
-    addresses      = ["10.17.3.${count.index}"]
+    addresses      = ["${cidrhost("${var.caasp_net_network}", 768 + count.index)}"]
```
ereslibre: Regarding the numbers, I think they are fine, but IMHO it would be even better if terraform supported bit shifting (not sure), so those numbers might be represented as shift operations: `1 << 8`, `2 << 8`, `3 << 8`. I think this is a bit more clear.

flavio: This is [not yet supported](https://www.terraform.io/docs/configuration/interpolation.html#math).

caasp-kvm/README.md
```diff
@@ -128,3 +128,89 @@ final octet incremented by 1:
...
+be swallowed by the VPN.
+At the same time you cannot pick networks like `172.16.0.0/13` or
+`172.17.0.0/24` because the first one is being used by flannel, while the second
+one is used by docker.
+A working subnet is `172.30.0.0/22`.
+
+Finally you have to add a route to the cluster on the remote machine. The
+easiest way is to use the following command:
+
+
```
ereslibre: Extra blank

caasp-kvm/README.md
```diff
@@ -128,3 +128,89 @@ final octet incremented by 1:
...
+```
+ip -4 route add 172.30.0.0/22 via 192.168.1.39
+```
+
```
ereslibre: Extra blank

caasp-kvm/README.md
```diff
@@ -128,3 +128,89 @@ final octet incremented by 1:
...
+You have to be really careful with the subnet you are going to use. The default
+one is conflicting with the SUSE R&D network. If your remote machine is
+connected to the corporate VPN all the requests made against your cluster will
+be swallowed by the VPN.
+At the same time you cannot pick networks like `172.16.0.0/13` or
+`172.17.0.0/24` because the first one is being used by flannel, while the second
+one is used by docker.
+A working subnet is `172.30.0.0/22`.
+
+Finally you have to add a route to the cluster on the remote machine. The
```
ereslibre: Instead of `remote machine` we could use `development machine` as it has been named before (so it's clear where the route should be added).

flavio:
@ereslibre: fixed the README as you pointed out.

PR #180, opened 2017-10-24
Fix rspec devenv

flavio:
Right now running `rspec` inside of the dashboard container will cause the production database to be wiped.

This change ~~(together with PR https://github.com/kubic-project/caasp-container-manifests/pull/132)~~ allows rspec to use the `velum_test` database.


flavio:
[#devenv create a test database for rspec](https://trello.com/c/dEW2zZMW/429-devenv-create-a-test-database-for-rspec)

flavio:
[#devenv create a test database for rspec](https://trello.com/c/dEW2zZMW/429-devenv-create-a-test-database-for-rspec)

caasp-kvm/README.md
```diff
@@ -128,3 +128,11 @@ final octet incremented by 1:
 * Admin node will have 10.17.1.0
 * First master node will have 10.17.2.0
 * First worker node will have 10.17.3.0
+
+## Running development tests against velum
+
+To run velum's rspec tests:
+
+* ssh into the admin node
```
MaximilianMeister: can we also execute another command on the mariadb container here to create the test db? this would likely decouple us from having to change the `setup-mariadb.sh` script in https://github.com/kubic-project/caasp-container-manifests/pull/132

flavio:
There's no more need to have the manifest file changed to always create a `velum_test` database.

This can be done on the admin node by running the following command:

```
docker run --rm -v /var/run/mysql/:/var/run/mysql/ -v /var/lib/misc/infra-secrets/:/infra-secrets:ro -v /usr/share/caasp-container-manifests/setup/mysql/setup-mysql.sh:/setup-mysql.sh:ro -e ENV=test --entrypoint /setup-mysql.sh sles12/mariadb:latest
```

This command has to be run once to create the `velum_test` database. It complains about the `CREATE USER` failing for `velum@localhost` and `salt@localhost` because they already exists. However everything is working as expected. The bogus error message can be solved once we update to the next release of mariadb because it adds support for the `CREATE USER IF EXISTS` statement.

In the meantime the real question is: how do we automate all of that? I could update the README.md to include the new docker command but this is not really user friendly.

I was thinking about creating a `run-rspec.sh` (or something similar) script and have it injected into the admin VM (we could add other helper scripts too, I've seen them around in some of our repos).

@ereslibre, @MaximilianMeister: what do you prefer?

caasp-kvm/README.md
```diff
@@ -128,3 +128,11 @@ final octet incremented by 1:
...
+* execute: `docker exec -ti $(docker ps | grep dashboard | awk {'print $1'}) entrypoint.sh bash -c "unset VELUM_DB_NAME; export RAILS_ENV=test; rspec"`
```
ereslibre: What about for making this simpler we change https://github.com/kubic-project/velum/blob/master/config/database.yml, so `test` database is always pointing to `velum_test`? This way we wouldn't have to `unset VELUM_DB_NAME`, no?

flavio: @ereslibre: done with https://github.com/kubic-project/velum/pull/372

flavio:
Everything can be automated by using this script:

```bash
#!/bin/bash

echo "Setup test database"
docker run --rm -v /var/run/mysql/:/var/run/mysql/ -v /var/lib/misc/infra-secrets/:/infra-secrets:ro -v /usr/share/caasp-container-manifests/setup/mysql/setup-mysql.sh:/setup-mysql.sh:ro -e ENV=test --entrypoint /setup-mysql.sh sles12/mariadb:latest

echo "Run rspec"
docker exec $(docker ps | grep dashboard | awk {'print $1'}) entrypoint.sh bash -c "unset VELUM_DB_NAME; export RAILS_ENV=test; rspec"
```

I could add it to this PR, I just don't know where to place it inside of this repo. Any idea?

caasp-kvm/tools/build-velum-image
```diff
@@ -50,11 +50,16 @@ build_fresh_image() {
...
+                      cp -r /srv/velum/vendor/bundle/ruby /var/lib/velum/"`
+  do_commit "$CONTAINER_BUILD"
+
+
+  local CONTAINER_BUILD=`docker run -v $VELUM_DIR:/srv/velum -d sles12/velum:development_wip \
+                    bash -c "
+                      cd /srv/velum && \
                       bundle config --local frozen 0 && \
                       bundle config --local build.nokogiri --use-system-libraries && \
                       bundle install --binstubs=/usr/local/bin --deployment --path=/var/lib/velum && \
```
ereslibre: I'm not sure about this entire commit (71fbba6467a5e23c262f2066c2d354fa368cb001). If you have local changes on the `Gemfile{,.lock}` you will never get a valid bundle, and when we mount the source code nothing will start as the `Gemfile.lock` won't match the bundled one.

I think we should keep this as it was, not sure what problem this was fixing.

ereslibre:
https://github.com/kubic-project/automation/pull/184 is ready to be reviewed and would supersede this PR. Please, give your input to know if that one is fine.

flavio:
Closing of favor of #184

PR #184, opened 2017-10-31
Fix rspec devenv

ereslibre:
This is a try to fix the rspec devenv in a slightly different way than proposed in https://github.com/kubic-project/automation/pull/180.

It's another PR just for the purpose of discussing both cleanly.

Depends on: https://github.com/kubic-project/velum/pull/378
Supersedes: https://github.com/kubic-project/automation/pull/180
Supersedes: https://github.com/kubic-project/velum/pull/372

### DIFF
--- a/caasp-kvm/README.md
+++ b/caasp-kvm/README.md
@@ -219,4 +219,9 @@ Now you can go to the workstation and use caasp-kvm as usual.
 To run velum's rspec tests:
 
 * ssh into the admin node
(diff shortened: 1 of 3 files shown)


ereslibre:
Adding `WIP` because salt is configured to access `velum_production` database. This means that if we manually deploy a cluster within this environment, the event processor won't get any events from salt (as they are reading/writing from/to different databases).

We should modify the salt-master config under this environment to use `velum_development`...

PR #185, opened 2017-11-02
WIP and RFC: bring back caasp kvm qa project

flavio:
The purpose of this PR is to fill the gap between `automation/caasp-kvm` and the original caasp-kvm project.

Without this PR `automation/caasp-kvm` creates caasp clusters where the admin node has development code injected. This is useful for development purposes, but it doesn't help when doing QA or debugging tasks.

With this PR the `caasp-kvm` script has been extended with a new flag: `--devenv`. By default the flag is set to false, leading to the quick creation of "vanilla" caasp clusters where the admin node doesn't have any development code injected.
On the opposite, when the `--devenv` flag is used, the cluster will be created in the old usual way.

Due to the lack of conditionals I resorted to the usage of a terraform module to describe the admin node. This has been done to reduce the code duplication.

This is a WIP and, most important of all, a RFC. If others are fine with this approach I'll extend the documentation of the project to cover this new use case.

### DIFF
--- a/caasp-kvm/caasp-kvm
+++ b/caasp-kvm/caasp-kvm
@@ -8,7 +8,7 @@ DIR="$( cd "$( dirname "$0" )" && pwd )"
 ACTION=
 RUN_BUILD=
 RUN_DESTROY=
-ENABLE_DEVENV=
+VANILLA=
 
 MASTERS=${CAASP_NUM_MASTERS:-1}
 WORKERS=${CAASP_NUM_WORKERS:-2}
@@ -39,7 +39,7 @@ Usage:
     -m|--masters <INT>     Number of masters to build (Default: CAASP_NUM_MASTERS=$MASTERS)
     -w|--workers <INT>     Number of workers to build (Default: CAASP_NUM_WORKERS=$WORKERS)
     -i|--image <STR>       Image to use (Default: CAASP_IMAGE=$IMAGE)
-    -d|--devenv            Enable devenv (Default: false)
+    --vanilla              Do not inject devenv code, use vanilla caasp (Default: false)
 
   * Destroying a cluster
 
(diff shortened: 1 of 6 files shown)


flavio:
Note well: right now the PR is based on https://github.com/kubic-project/automation/pull/173. I'll rebase it once https://github.com/kubic-project/automation/pull/173 gets merged.

caasp-kvm/caasp-kvm
```diff
@@ -38,6 +39,7 @@ Usage:
     -m|--masters <INT>     Number of masters to build (Default: CAASP_NUM_MASTERS=$MASTERS)
     -w|--workers <INT>     Number of workers to build (Default: CAASP_NUM_WORKERS=$WORKERS)
     -i|--image <STR>       Image to use (Default: CAASP_IMAGE=$IMAGE)
+    -d|--devenv            Enable devenv (Default: false)
```
kiall: I think we should do the opposite here, and instead allow for disabling-devenv - given we use for dev way more often than for validation scenarios, it makes sense to me to default this to enabled...

flavio:
@kiall: I've changed the behaviour to follow your recommendation. Now the devenv is used by default.

bs-github:
I have done basically the same thing in a different way https://github.com/bs-suse/automation/commit/1fad5ac295db642b8cad6cd4178fde00bb0e8ce8 before I found the branch from @flavio and this pull request. 
I don't think my implementation is clean enough to merge (in regards of variable names and stuff), but I'd like you to consider a less complex way and less code duplication.
My reason to go for the erb template was having one file with both definitions (with and without injected code).

flavio:
Yep, using erb makes things nicer.

We already did that with the old [https://github.com/kubic-project/terraform/tree/release-2.0/terraform](terraform) project and it turned into a gigantic pile of ruby erb files doing tons of things.

I'm fine in going the erb way as long as we avoid going the mess we had in the past. What do the others think?

bs-suse:
I'd like to get this or PR #216 in to make use of it in QA testing.

flavio:
Closed in favor of https://github.com/kubic-project/automation/pull/216. There are a couple of minor cleanups that I'll carry over from this patch.

PR #191, opened 2017-11-24
Move manifests into a subdirectory

kiall:
This allows us to remove the public.yaml / private.yaml hardcodes, which in
turn allows us to split the public / private pods into more specific pods
matching the typical patterns used to deploy workloads in K8S.

See also - https://github.com/kubic-project/caasp-container-manifests/pull/135

### DIFF
--- a/caasp-kvm/caasp-kvm
+++ b/caasp-kvm/caasp-kvm
@@ -211,7 +211,12 @@ build() {
     cp -r $CAASP_MANIFESTS_DIR/* "$injected/"
 
     log "Patching Container Manifests"
-    $DIR/tools/fix-kubelet-manifest -o $injected/public.yaml $injected/public.yaml
+    # TODO: Remove the if/else once https://github.com/kubic-project/caasp-container-manifests/pull/135 is merged
+    if [ -f "$injected/manifests/public.yaml" ]; then
+      $DIR/tools/fix-kubelet-manifest -o $injected/manifests/public.yaml $injected/manifests/public.yaml
+    else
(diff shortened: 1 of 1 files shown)


kiall:
> LGTM (we should merge the final version only)

What do you mean by final version? You mean without the if/else? That's actually intentional and intended to be merged... it allows CI to pass, and means no force merges - something we need to get better at not doing :P

ereslibre:
> What do you mean by final version? You mean without the if/else? That's actually intentional and intended to be merged... it allows CI to pass, and means no force merges - something we need to get better at not doing :P

My idea was to merge https://github.com/kubic-project/caasp-container-manifests/pull/135 (we still need to force merge that...). When that happens, this one gets the `if/else` removed (and should be green) and that patch is merged with this PR (without the `if/else`).

This one will be green also (also when we remove the `if/else`) when https://github.com/kubic-project/caasp-container-manifests/pull/135 gets merged, but we will blindly merge https://github.com/kubic-project/caasp-container-manifests/pull/135 anyway :)

kiall:
> My idea was to merge kubic-project/caasp-container-manifests#135 (we still need to force merge that...). 

We don't need to force merge anything :) Merge this, rerun CI on https://github.com/kubic-project/caasp-container-manifests/pull/135 - it'll pass - then merge it, and remove the if/else from here.

ereslibre:
> We don't need to force merge anything :) Merge this, rerun CI on kubic-project/caasp-container-manifests#135 - it'll pass - then merge it, and remove the if/else from here.

You're right. +1 for merge this as it is.

PR #192, opened 2017-11-27
Use admin IP (vs. hostname) in velum-bootstrap

BlaineEXE:
For environments where there is not DNS resolution to the admin by
hostname, the external IP address must be used. This will still support
the DNS-enabled environment, so simply replace the hostname reference
with the external IP address reference.


ereslibre:
> However - CaaSP expects a DNS name, and I think will not function correctly with an IP instead of DNS name here.

I can't think of a reason why it wouldn't work at the moment. We have always allowed IP addresses and FQDN's on all fields except on the "Kubernetes External FQDN" name where placing an IP address would not work as this value was always added as `DNS:` value on the certificate. This PR also fixed this: https://github.com/kubic-project/salt/pull/311

(That being said, we should always encourage specially on this field to use FQDN's)

BlaineEXE:
@kiall @ereslibre: Should I look instead into getting DNS working with the Heat tooling instead of this?

BlaineEXE:
This doesn't seem to be a fix as I thought it was. Closing and can be reopened later if it makes sense.

kiall:
Looks like you forgot to hit close :) Reopen if needs be!

PR #193, opened 2017-11-27
Add 'kubernetesExternalHost' to Heat environment

BlaineEXE:
For velum-bootstrap tooling to work, it now requires a
'kubernetesExternalHost' field to be present. Add this into the
environment.json file generation in the OpenStack Heat tooling.
Trivially select the first master as the external kubernetes host.

### DIFF
--- a/caasp-openstack-heat/tools/generate-environment
+++ b/caasp-openstack-heat/tools/generate-environment
@@ -25,6 +25,9 @@ INDEX=$((INDEX + 1))
 
 # Iterate all the masters
 MASTER_IDS=$(set -o pipefail; echo "$STACK_RESOURCES" | jq -r '.[] | select(.stack_name | contains("-masters-")) | select(.resource_type == "OS::Nova::Server") | .physical_resource_id')
+K8S_MASTER_NOVA_JSON=$(set -o pipefail; openstack server show -f json ${MASTER_IDS%%$'\n'*}) # Pick the first master to be the K8S public endpoint
+K8S_MASTER_PUBLIC_IPV4=$(set -o pipefail; echo "$K8S_MASTER_NOVA_JSON" | jq -r ".addresses | split (\"=\") | .[1] | split(\",\") | .[1] | gsub(\" \"; \"\")")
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
can't say if it works, as i havent tested it, but i remember having had the same issue once. i wonder if we just need to sync the code from the other `generate-environment` script: https://github.com/kubic-project/automation/blob/0893ec2f5400d2308ceaa0fef32b30f263cb4cb3/caasp-kvm/tools/generate-environment#L41 where we set `kubernetesExternalHost` as well

kiall:
I *think* CaaSP actually needs a DNS name for this field, rather than a IP - otherwise bad things will happen... Maybe we should just add some OpenStack DNS resources to the heat templates and get real+working DNS...

ereslibre:
This comment also applies here: https://github.com/kubic-project/automation/pull/192#issuecomment-347548649

BlaineEXE:
Would it be appropriate to move `generate-environment` to the `misc-tools/` dir and use that in all tooling?

BlaineEXE:
It seems like the CI failure here may be related to the DNS/IP issues. The velum-bootstrap runs properly, but the cluster doesn't bootstrap successfully.

MaximilianMeister:
@BlaineEXE sorry for the late answer, yes the 2 create-environment scripts are too different, and project specific...

however the bootstrap failure is not related to DNS/IP as it is green in other scenarios, and your code change here shouldnt even influence the jenkins run anyways. maybe a random failure... i just retriggered

BlaineEXE:
@MaximilianMeister I also don't see how this could affect the jenkins run. It still looks like it's failing. Could there be something in a recent PR that's affected the CI?

MaximilianMeister:
> @MaximilianMeister I also don't see how this could affect the jenkins run. It still looks like it's failing. Could there be something in a recent PR that's affected the CI?

good question, at least the error is different now. bootstrap was fine. maybe your branch is too old. could you rebase to upstream/master ?

kiall:
> good question, at least the error is different now. bootstrap was fine. maybe your branch is too old. could you rebase to upstream/master ?

Yea, a rebase is needed to pass due to other changes

BlaineEXE:
Check is passing now.

PR #194, opened 2017-11-28
OpenStack Heat - Enable use of OpenStack DNS

kiall:
WIP - Untested.


mjura:
Merging this PR we are putting strong requirement for OpenStack DNS Service which is still not common service for OpenStack deployments. Maybe we should add some check if OpenStack DNS Service is available first and then use it.

MalloZup:
closing since never used and improved since 1 year, we don't need it. 

Thank you

PR #195, opened 2017-11-29
CaaSP OS Heat: Add optional disks to workers

BlaineEXE:
For building applications on CaaSP that leverage host storage, add
ability to specify in the heat environment file that workers can have a
configurable number of disks added with a configurable size. Default is
to add zero disks, and default disk size is 10 GB.

### DIFF
--- a/caasp-openstack-heat/caasp-stack-worker.yaml
+++ b/caasp-openstack-heat/caasp-stack-worker.yaml
@@ -39,6 +39,14 @@ parameters:
   admin_node_ip:
     type: string
     description: Admin Node IP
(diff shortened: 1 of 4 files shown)


BlaineEXE:
@kiall, could you take a look?

BlaineEXE:
Also, would it be possible to add me as a kubic-project member so jenkins will automatically test my PRs? As it stands, I have to do a manual step to allow a CI test.

kiall:
@BlaineEXE I think I've set things up so that, once you accept the invitation, jobs will start automatically for you.

PR #196, opened 2017-11-30
CaaSP OS Heat: Allow all nodes to ssh each other

BlaineEXE:
Add rsa private and public keys to the cloud config of all nodes to
allow them to passwordless ssh to each other. This makes it easier to
move around the cluster for development.


caasp-openstack-heat/caasp-stack-worker.yaml
```diff
@@ -78,6 +78,44 @@ resources:
...
+                yKx3AoGAOp6PfnqgxrbFp6O46NcRXtWNLKKulsi/m1mmPAuDsHQJPYKjpsi9iB8a
+                jSM46H8miC4aqfCln5pDHy0kG7vG//0R0SnByzy5KW+KbAvv2eDU2WO8b9SMXkmH
+                tGE3GP+Pj4qHaTLacHos6s6W2UOSMWxU1cWUGikgpVXdQngx9bY=
+                -----END RSA PRIVATE KEY-----
+
+              rsa_public: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC2G7k0zGAjd+0LzhbPcGLkdJrJ/LbLrFxtXe+LPAkrphizfRxdZpSC7Dvr5Vewrkd/kfYObiDc6v23DHxzcilVC2HGLQUNeUer/YE1mL4lnXC1M3cb4eU+vJ/Gyr9XVOOReDRDBCwouaL7IzgYNCsm0O5v2z/w9ugnRLryUY180/oIGeE/aOI1HRh6YOsIn7R3Rv55y8CYSqsbmlHWiDC6iZICZtvYLYmUmCgPX2Fg2eT+aRbAStUcUERm8h246fs1KxywdHHI/6o3E1NNIPIQ0LdzIn5aWvTCd6D511L4rf/k5zbdw/Gql0AygHBR/wnngB5gSDERLKfigzeIlCKf insecure-key
+
+            # RSA keys above are put into /etc/ssh; copy them to /root/.ssh/
+            runcmd:
+              - mkdir --parents /root/.ssh
```
MalloZup: Is better to use standard salt.states instead of run.cmd

MalloZup: https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html  for files

BlaineEXE: If someone would like to propose those changes, that would be great. I don't know how to do this using salt.states.

mjura:
Sorry but this looks like bad solution:
1. Hardcoded private key in Heat template is totally wrong. This have to generated for each deployment and uploaded to OpenStack, e.g.:
```
nova keypair-add --pub-key ~/.ssh/id_rsa.pub default
```
2. Heat template has already `OS::Nova::Server` resource option called `key_name: {get_param: ssh_key_name}`

mjura:
Please ping on IRC and we can work on this together

BlaineEXE:
@mjura This PR is not to make it so that the dev machine can access the cluster; this is already done, as you mention. It is to allow each node in the cluster to access every other node in the cluster. The private key is hardcoded because it is a direct copy of the private key found in the `misc_files` directory, which is what is given to heat by the tooling, as far as I can tell. I have found no way in openstack to add a private key to the root user of all nodes from a file except to use cloud init.

BlaineEXE:
I wrote a script that just copies the keys I need without having to use cloud-config.

PR #199, opened 2017-12-01
Load images during Velum bootstrap

kiall:
This allows for more accurate screenshots during CI runs.


velum-bootstrap/spec/spec_helper.rb
```diff
@@ -31,7 +31,7 @@ def admin_minion
     js_errors:         false,
     phantomjs_options: [
       "--proxy-type=none",
-      "--load-images=no"
+      "--load-images=yes"
```
MaximilianMeister: [default is yes](http://phantomjs.org/api/webpage/property/settings.html) so we might also drop the line completely

PR #200, opened 2017-12-08
Heat: WIP: Fix masters/workers not having internet access

BlaineEXE:
Workers did not have network access to pull images from quay.io or
docker.io. Adding a public interface allows them to do so.


BlaineEXE:
I'm not sure if this is an issue just with my environment. This seemed like the fastest way for me to give my minions internet access so they can pull external containers, but I'm not sure it's the best solution.

I think I can also configure the nodes such that the workers will use the admin or the masters as proxy access to the internet. I'd like to get some feedback as to whether this solution is desirable or if it would be better to try a different alternative.

BlaineEXE:
I'm actually figuring out now that although my workers have IP addresses listed, they still don't have internet access. I can't ping anything...

mjura:
You don't have to assign Public IPs to Worker nodes to get access to internet. Worker nodes shouldn't have Public IP's. Please configure internal network properly with router gateway and it will be working fine.

mjura:
Please ping me on IRC and I can help you.

BlaineEXE:
@mjura The environment is created by Heat, which configures an external router automatically. As per my "Update 2" comment, the issue seems to be with iptables. When I mangled iptables and allowed all traffic, the nodes were able to pull containers from the internet. 

This may imply what you have said that if the internal router is configured, the nodes shouldn't need IPs, but I still need to resolve the issue of the node firewalls blocking access.

BlaineEXE:
I found the source of the issue in my case. I set my `heat-environment.yaml` file to set
```yaml
parameters:
  dns_nameserver: 8.8.8.8
```

The default value for the `dns_nameserver` value is 172.24.0.2. I am assuming the 172.24.0.2 value works for the CaaSP team. If this is the case, I would like to change the default value in the Heat template to 8.8.8.8, and the default Heat automation should still work for the CaaSP team, and for others such as myself, they may find it more clear that this value needn't have any relation to the OpenStack network.

BlaineEXE:
Do you have any thoughts, @kiall?

kiall:
@BlaineEXE setting it to a public DNS service like Google's 8.8.8.8 means we won't be able to resolve any SUSE internal names... (e.g. internal registry mirrors etc) - or $company internal when ran by others.

172.24.0.2 should be the IP OpenStack assigns to the Neutron managed dnsmasq instance (assuming you haven't also changed the network ranges, in which case, the neutron DNS instance will be the second IP in that range and both values need to be changed together).

If that's not working, then your OpenStack cloud is likely misconfigured and missing the `dnsmasq_dns_servers` options in Neutron's `dhcp_agent.ini` config - and this option is there to let you configure it to whatever happens to work for your environment.

BlaineEXE:
Ohhhhhh... So this is something I can take up with ECP. Roger.

BlaineEXE:
I think I have confirmed that this is an ECP issue, though I don't know where it exists. I am closing this PR.

PR #201, opened 2017-12-11
Automatic dns entry configuration for admin and master nodes

ellisab:
With this change I am able to create automatically dns entries for admin and master nodes.
It introduces a new variable dns_domain


mjura:
Is required Openstack Designate DNSaaS for this? Change looks fine but I doubt how does it work without Openstack Designate

mjura:
As I see in Terraform documentation OpenStack DNS service is needed 
https://www.terraform.io/docs/providers/openstack/r/dns_zone_v2.html

I'm not sure if we should merge it

pgonin:
Would that work with a count variable ?
Set to 0 by default (if no Designate available)

PR #202, opened 2017-12-18
enable adding custom repos to the master/workers

MaximilianMeister:

### DIFF
--- a/caasp-devenv
+++ b/caasp-devenv
@@ -32,6 +32,7 @@ DISABLE_MELTDOWN_SPECTRE=${CAASP_DISABLE_MELTDOWN_SPECTRE:-}
 PROXY=${CAASP_HTTP_PROXY:-}
 PARALLELISM=${CAASP_PARALLELISM:-1}
 ENABLE_TILLER=${CAASP_ENABLE_TILLER:-false}
+EXTRA_REPO=${CAASP_EXTRA_REPO:-}
 
 # base distribution used for setting up dependencies, and the packages to install
 DIST_BASE="opensuse.org/repositories/Virtualization:/containers"
@@ -57,7 +58,7 @@ Usage:
 
   * Setup your workstation
 
-    --setup                  Install Dev Env Dependencies
+    --setup                     Install Dev Env Dependencies
 
   * Building a cluster
 
@@ -75,16 +76,20 @@ Usage:
     Velum Username: test@test.com
     Velum Password: password
 
-    -B|--bootstrap           Bootstrap CaaSP cluster
+    -B|--bootstrap              Bootstrap CaaSP cluster
 
   * Testing a cluster
 
-    -T|--testinfra           Run testinfra tests
+    -T|--testinfra              Run testinfra tests
 
   * Common options
 
-    -p|--parallelism         Set terraform parallelism (Default: CAASP_PARALLELISM)
-    -P|--proxy               Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+    -p|--parallelism            Set terraform parallelism (Default: CAASP_PARALLELISM)
+    -P|--proxy                  Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+
+  * Advanced Options
+
+    --extra-repo <STR>          URL of a custom repository on the master(s)/worker(s) (Default: CAASP_EXTRA_REPO)
 
   * Examples:
 
@@ -95,6 +100,10 @@ Usage:
   Bootstrap and Test a pre-made cluster
 
   $0 --bootstrap --testinfra
+
+  Build a cluster and add a custom repository on the master/worker(s)
+
+  $0 --build -m 1 -w 2 --extra-repo https://download.opensuse.org/repositories/devel:/CaaSP:/Head:/ControllerNode:/TestUpdates
 
 USAGE
 )
@@ -158,6 +167,10 @@ while [[ $# > 0 ]] ; do
       PROXY="$2"
       shift
       ;;
+    --extra-repo)
+      EXTRA_REPO="$2"
+      shift
+      ;;
     -h|--help)
       usage
       ;;
@@ -168,10 +181,11 @@ done
 ##############################################################
 
 CAASP_KVM_ARGS="-m $MASTERS -w $WORKERS --image $IMAGE"
-[ -n "$PARALLELISM"              ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --parallelism=$PARALLELISM"
-[ -n "$PROXY"                    ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --proxy $PROXY"
-[ -n "$VANILLA"                  ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --vanilla"
+[ -n "$PARALLELISM" ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --parallelism=$PARALLELISM"
+[ -n "$PROXY"       ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --proxy $PROXY"
+[ -n "$VANILLA"     ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --vanilla"
 [ -n "$DISABLE_MELTDOWN_SPECTRE" ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --disable-meltdown-spectre-fixes"
+[ -n "$EXTRA_REPO"  ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --extra-repo $EXTRA_REPO"
 
 export ENVIRONMENT="$DIR/caasp-kvm/environment.json"
 
--- a/caasp-kvm/README.md
+++ b/caasp-kvm/README.md
@@ -95,18 +95,18 @@ sudo virsh pool-start default
 
       * Destroying a cluster
 
-        -d|--destroy           Run the CaaSP KVM Destroy Step
+        -d|--destroy                Run the CaaSP KVM Destroy Step
 
       * Common options
 
-        -p|--parallelism       Set terraform parallelism (Default: CAASP_PARALLELISM)
-        -P|--proxy             Set HTTP proxy (Default: CAASP_HTTP_PROXY)
+        -p|--parallelism            Set terraform parallelism (Default: CAASP_PARALLELISM)
+        -P|--proxy                  Set HTTP proxy (Default: CAASP_HTTP_PROXY)
 
       * Local git checkouts
 
-         --salt-dir <DIR>      the Salt repo checkout (Default: CAASP_SALT_DIR)
-         --manifests-dir <DIR> the manifests repo checkout (Default: CAASP_MANIFESTS_DIR)
-         --velum-dir <DIR>     the Velum repo checkout (Default: CAASP_VELUM_DIR)
+         --salt-dir <DIR>           the Salt repo checkout (Default: CAASP_SALT_DIR)
+         --manifests-dir <DIR>      the manifests repo checkout (Default: CAASP_MANIFESTS_DIR)
+         --velum-dir <DIR>          the Velum repo checkout (Default: CAASP_VELUM_DIR)
 
       * Advanced Options
 
@@ -116,6 +116,7 @@ sudo virsh pool-start default
         --worker-ram <INT>     CPUs to allocate to master node(s) (Default: CAASP_MASTER_CPU=2)
         --master-cpu <INT>     RAM to allocate to worker node(s) (Default: CAASP_WORKER_RAM=2048)
         --worker-cpu <INT>     CPUs to allocate to worker node(s) (Default: CAASP_WORKER_CPU=2)
+        --extra-repo <STR>     URL of a custom repository on the master(s)/worker(s) (Default: CAASP_EXTRA_REPO)
 
       * Examples:
 
@@ -126,6 +127,10 @@ sudo virsh pool-start default
       Build a 1 master, 2 worker cluster using the latest staging A image
 
       ./caasp-kvm --build -m 1 -w 2 --image channel://staging_a
+
+      Build a 1 master, 2 worker cluster and add a custom repository on the master(s)/worker(s)
+
+      ./caasp-kvm --build -m 1 -w 2 --extra-repo https://download.opensuse.org/repositories/devel:/CaaSP:/Head:/ControllerNode:/TestUpdates
 
       Destroy a cluster
 
@@ -141,6 +146,26 @@ We maintain packages for both of them inside of the
 [Virtualization:containers](https://build.opensuse.org/project/show/Virtualization:containers)
 project on OBS.
 
+## Using a cluster with custom rpm packages from a repository
+
+In the devenv we can already inject code from a few components (e.g. velum, salt, manifests)
+
+But to test e.g. a different docker image, or another version of kubernetes you can add a custom repository url to the commandline and this repo will then be added to either the master(s), the worker(s) or both:
+
+    ./caasp-kvm --build -m 1 -w 2 --extra-repo https://download.opensuse.org/repositories/devel:/CaaSP:/Head:/ControllerNode:/TestUpdates
+
+After the cluster is up and running, you can either head over to the velum user interface and trigger the update (after transactional-update has run), or install the new packages on master/worker nodes via:
+
+    # on the host
+    RESOLV_CONF=$(< /etc/resolv.conf) transactional-update shell
+    # in the transactional-update shell
+    echo "$RESOLV_CONF" > /etc/resolv.conf
+    zypper -n dup --allow-vendor-change
+    exit
+    # on the host
+    reboot
+
+Writing again the resolv.conf is necessary as the there is no overlayfs in the chroot environment of the transactional-update shell
 
 ## Using a cluster from the hypervisor node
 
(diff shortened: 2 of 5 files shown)


caasp-devenv
```diff
@@ -29,6 +29,8 @@ WORKERS=${CAASP_NUM_WORKERS:-2}
 IMAGE=${CAASP_IMAGE:-channel://devel}
 PROXY=${CAASP_HTTP_PROXY:-}
 PARALLELISM=${CAASP_PARALLELISM:-1}
+MASTER_PTF_REPO_URL=${MASTER_PTF_REPO_URL:-0}
+WORKER_PTF_REPO_URL=${WORKER_PTF_REPO_URL:-0}
```
ereslibre: Why not default them to `${MASTER_PTF_REPO_URL:-}`. Being a URL I think it makes more sense to default them to nothing instead of 0.

caasp-kvm/README.md
```diff
@@ -106,10 +108,35 @@ Clone these repositories:
 
       ./caasp-kvm --build -m 1 -w 2 --image channel://staging_a
 
+      Build a 1 master, 2 worker cluster and add a custom repository on the master/worker(s)
+
+      ./caasp-kvm --build -m 1 -w 2 --master-ptf-repo-url http://download.suse.de/ibs/Devel:/CASP:/Head:/ControllerNode:/TestUpdate/standard/ --worker-ptf-repo-url http://download.suse.de/ibs/Devel:/CASP:/Head:/ControllerNode:/TestUpdate/standard/
```
ereslibre: As a side note, since on CaaSP we release the same image for everything I think we will be duplicating this arguments more than anything when calling, maybe we can just add `--override-ptf-repo-url` or something like that, but it applies to masters as well as workers.

mjura: Can you provide please some OBS example repository? It will be better from opensource perspective

caasp-kvm/README.md
```diff
@@ -106,10 +108,35 @@ Clone these repositories:
...
 
       ./caasp-kvm --destroy
 
+## Using a cluster with custom rpm packages from a repository
+
+In the devenv we can already inject code from a few components (e.g. velum, salt, manifests)
+
+But to test e.g. a different docker image, or another version of kubernetes you can add a custom repository url to the commandline and this repo will then be added to either the master(s), the worker(s) or both:
+
+    ./caasp-kvm --build -m 1 -w 2 --master-ptf-repo-url http://download.suse.de/ibs/Devel:/CASP:/Head:/ControllerNode:/TestUpdate/standard/ --worker-ptf-repo-url http://download.suse.de/ibs/Devel:/CASP:/Head:/ControllerNode:/TestUpdate/standard/
```
mjura: Can you provide please some OBS example repository? It will be better from opensource perspective

caasp-kvm/README.md
```diff
@@ -106,10 +108,35 @@ Clone these repositories:
...
+After the cluster is up and running, you can install the new packages on master/worker nodes via:
+
+    # on the host
+    RESOLV_CONF=$(< /etc/resolv.conf) transactional-update shell
+    # in the transactional-update shell
+    echo "$RESOLV_CONF" > /etc/resolv.conf
+    zypper -n dup --allow-vendor-change
+    exit
+    # on the host
+    reboot
```
ereslibre: Maybe it makes sense to modify the cloud-init file so it includes `runcmd` and run this in an automated way, so there's no need to perform any actions manually.

caasp-devenv
```diff
@@ -148,6 +167,8 @@ done
 CAASP_KVM_ARGS="-m $MASTERS -w $WORKERS --image $IMAGE"
 [ -n "$PARALLELISM" ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --parallelism=$PARALLELISM"
 [ -n "$PROXY"       ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --proxy $PROXY"
+[ "$MASTER_PTF_REPO_URL" != 0 ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --master-ptf-repo-url $MASTER_PTF_REPO_URL"
```
mjura: This will be always true, please change it to:
```
[ -n "$MASTER_PTF_REPO_URL" ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS ....
```

caasp-kvm/caasp-kvm
```diff
@@ -26,6 +26,9 @@ MASTER_CPU=${CAASP_MASTER_CPU:-2}
 WORKER_RAM=${CAASP_WORKER_RAM:-2048}
 WORKER_CPU=${CAASP_WORKER_CPU:-2}
 
+MASTER_PTF_REPO_URL=${MASTER_PTF_REPO_URL:-0}
+WORKER_PTF_REPO_URL=${WORKER_PTF_REPO_URL:-0}
```
ereslibre: Same here with the defaults.

caasp-devenv
```diff
@@ -148,6 +167,8 @@ done
...
+[ "$WORKER_PTF_REPO_URL" != 0 ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --worker-ptf-repo-url $WORKER_PTF_REPO_URL"
```
mjura: This will be always true, please change it to:
```
[ -n "$WORKER_PTF_REPO_URL" ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS ....
```

caasp-devenv
mjura: +1

ereslibre: Yup, I meant both.

caasp-devenv
```diff
@@ -47,31 +49,36 @@ Usage:
...
 
-    -p|--parallelism         Set terraform parallelism (Default: CAASP_PARALLELISM)
-    -P|--proxy               Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+    -p|--parallelism            Set terraform parallelism (Default: CAASP_PARALLELISM)
+    -P|--proxy                  Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+
+  * Advanced Options
+
+    --master-ptf-repo-url <STR> URL of a custom repository on the master
+    --worker-ptf-repo-url <STR> URL of a custom repository on the worker
```
mjura: Just wondering if we need two different repositories, probably one option `-R` and `--repo` should be enough

caasp-kvm/README.md
MaximilianMeister: wouldn't do this as this will run on CI as well, and the reboot will cause disruption. even if we make the reboot conditional, when no dup is necessary which is the normal case in CI, where we dont provide any ptf url. in case there is an update in a repo just at that time, it will be published in the repo and it will trigger a reboot, so i am not sure. ideally we wouldn't mount any repo but it is not possible to have conditions in the cloud-init config

caasp-kvm/README.md
MaximilianMeister: we could just use one argument, if we dont have any use case for testing a custom package only either on master or worker. is there any usecase?

caasp-kvm/README.md
MaximilianMeister: sure!

caasp-devenv
MaximilianMeister: for that i need to change the default, as @ereslibre mentioned. i can do that

mjura:
needs to be rebased

caasp-devenv
```diff
@@ -52,31 +53,35 @@ Usage:
...
   * Common options
 
-    -p|--parallelism         Set terraform parallelism (Default: CAASP_PARALLELISM)
-    -P|--proxy               Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+    -p|--parallelism            Set terraform parallelism (Default: CAASP_PARALLELISM)
+    -P|--proxy                  Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+
+  * Advanced Options
+
+    --devel-repo-url <STR>      URL of a custom repository on the master(s)/worker(s)
```
kiall: `--extra-repo` would be clearer - since `devel` might make poeple think you can't point it at a maintaince update repo etc.

caasp-devenv
```diff
@@ -141,6 +150,10 @@ while [[ $# > 0 ]] ; do
       PROXY="$2"
       shift
       ;;
+    --devel-repo-url)
```
kiall: I wonder if we need to support passing in multiple repos? i.e. building a bash array here instead of setting a string variable. This would allow for "Build 2.0 cluster, Add released updates for 2.0 repo, add my test package repo" style scenarios.

caasp-kvm/README.md
```diff
@@ -143,6 +147,26 @@ We maintain packages for both of them inside of the
...
 
+## Using a cluster with custom rpm packages from a repository
+
+In the devenv we can already inject code from a few components (e.g. velum, salt, manifests)
+
+But to test e.g. a different docker image, or another version of kubernetes you can add a custom repository url to the commandline and this repo will then be added to either the master(s), the worker(s) or both:
+
+    ./caasp-kvm --build -m 1 -w 2 --devel-repo-url https://download.opensuse.org/repositories/devel:/CaaSP:/Head:/ControllerNode:/TestUpdates
+
+After the cluster is up and running, you can install the new packages on master/worker nodes via:
```
kiall: Maybe add a note about how to do the upgrade to the new packages with Velum, then commandline /manual for special cases?

MaximilianMeister: fixed

caasp-devenv
MaximilianMeister: fixed

caasp-devenv
```diff
@@ -30,6 +30,7 @@ IMAGE=${CAASP_IMAGE:-channel://devel}
 PROXY=${CAASP_HTTP_PROXY:-}
 PARALLELISM=${CAASP_PARALLELISM:-1}
 ENABLE_TILLER=${CAASP_ENABLE_TILLER:-false}
+EXTRA_REPO=${EXTRA_REPO:-}
```
kiall: This should be:

    EXTRA_REPO=${CAASP_EXTRA_REPO:-}

we do this so all env vars we consume are obvious - `env | grep CAASP_` will show them all (Yea, it's not perfect and some have slipped though.. still)

caasp-devenv
```diff
@@ -69,16 +70,20 @@ Usage:
...
   * Common options
 
-    -p|--parallelism         Set terraform parallelism (Default: CAASP_PARALLELISM)
-    -P|--proxy               Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+    -p|--parallelism            Set terraform parallelism (Default: CAASP_PARALLELISM)
+    -P|--proxy                  Set HTTP Proxy (Default: CAASP_HTTP_PROXY)
+
+  * Advanced Options
+
+    --extra-repo <STR>      URL of a custom repository on the master(s)/worker(s)
```
kiall: Add to the end so the env var is doc'd:

    (Default: CAASP_EXTRA_REPO)

caasp-kvm/README.md
```diff
@@ -116,6 +116,7 @@ sudo virsh pool-start default
         --worker-ram <INT>     CPUs to allocate to master node(s) (Default: CAASP_MASTER_CPU=2)
         --master-cpu <INT>     RAM to allocate to worker node(s) (Default: CAASP_WORKER_RAM=2048)
         --worker-cpu <INT>     CPUs to allocate to worker node(s) (Default: CAASP_WORKER_CPU=2)
+        --extra-repo <STR>      URL of a custom repository on the master(s)/worker(s) (Default: EXTRA_REPO=http://download.suse.de/ibs/Devel:/CASP:/Head:/ControllerNode/standard)
```
kiall: Looks like this is out of date and has a wrong default listed (thats not the default :))

caasp-kvm/README.md
kiall: Do we need these instructions? People should really be using the velum upgrade process to suck these in?

caasp-kvm/cloud-init/master.cfg.tpl
```diff
@@ -29,3 +29,9 @@ bootcmd:
   - ip link set dev eth0 mtu 1400
 
 final_message: "The system is finally up, after $UPTIME seconds"
+
+# add custom repo
```
kiall: I'd still love to see one of my original comments addressed :)

> Ideally, we could move the repo injection from cloud-init to a small script in misc-tools/ that reads env.json, logs into each server and installs the repo - caasp-kvm / caasp-devenv keep the flag to use it, and just call the script once they see the flag, after terraform has come up.

This allows the injection code to work for KVM, OpenStack, Bare Metal, etc

caasp-kvm/README.md
MaximilianMeister: this is a bit hacky, and is actually the default, as you can see in https://github.com/kubic-project/automation/pull/202/files#diff-520506dc0b210b0a33c7ab69ba7cac60R230

caasp-kvm/cloud-init/master.cfg.tpl
MaximilianMeister: ok, makes sense. although i like more the cloud-init approach as it feels more integrated, but if it doesnt work in all environments it makes sense to do it via external script

MaximilianMeister:
@kiall i have added an external script in the latest commit to add the repos, please have a look, so i can squash everything in the end

caasp-kvm/README.md
```diff
@@ -116,6 +116,7 @@ sudo virsh pool-start default
         --worker-ram <INT>     CPUs to allocate to master node(s) (Default: CAASP_MASTER_CPU=2)
         --master-cpu <INT>     RAM to allocate to worker node(s) (Default: CAASP_WORKER_RAM=2048)
         --worker-cpu <INT>     CPUs to allocate to worker node(s) (Default: CAASP_WORKER_CPU=2)
+        --extra-repo <STR>     URL of a custom repository on the master(s)/worker(s) (Default: CAASP_EXTRA_REPO=http://download.suse.de/ibs/Devel:/CASP:/Head:/ControllerNode/standard)
```
kiall: Should be "(Default: CAASP_EXTRA_REPO)"

caasp-devenv
```diff
@@ -168,10 +181,11 @@ done
...
 
 CAASP_KVM_ARGS="-m $MASTERS -w $WORKERS --image $IMAGE"
-[ -n "$PARALLELISM"              ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --parallelism=$PARALLELISM"
-[ -n "$PROXY"                    ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --proxy $PROXY"
-[ -n "$VANILLA"                  ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --vanilla"
+[ -n "$PARALLELISM" ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --parallelism=$PARALLELISM"
+[ -n "$PROXY"       ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --proxy $PROXY"
+[ -n "$VANILLA"     ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --vanilla"
 [ -n "$DISABLE_MELTDOWN_SPECTRE" ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --disable-meltdown-spectre-fixes"
+[ -n "$EXTRA_REPO"  ] && export CAASP_EXTRA_REPO=$EXTRA_REPO
```
kiall: We deliberately don't export variables to avoid "hidden" interdependences. If something else needs the value, it should be explicitly passed to that other thing so it's visible.

In this case, we should be doing this:

    [ -n "$EXTRA_REPO"  ] && CAASP_KVM_ARGS="$CAASP_KVM_ARGS --extra-repo $EXTRA_REPO"

caasp-kvm/caasp-kvm
```diff
@@ -48,27 +50,28 @@ Usage:
...
-    --worker-ram <INT>     CPUs to allocate to master node(s) (Default: CAASP_MASTER_CPU=$MASTER_CPU)
-    --master-cpu <INT>     RAM to allocate to worker node(s) (Default: CAASP_WORKER_RAM=$WORKER_RAM)
-    --worker-cpu <INT>     CPUs to allocate to worker node(s) (Default: CAASP_WORKER_CPU=$WORKER_CPU)
+    --admin-ram <INT>           RAM to allocate to admin node (Default: CAASP_ADMIN_RAM=$ADMIN_RAM)
+    --admin-cpu <INT>           CPUs to allocate to admin node (Default: CAASP_ADMIN_CPU=$ADMIN_CPU)
+    --master-ram <INT>          RAM to allocate to master node(s) (Default: CAASP_MASTER_RAM=$MASTER_RAM)
+    --worker-ram <INT>          CPUs to allocate to master node(s) (Default: CAASP_MASTER_CPU=$MASTER_CPU)
+    --master-cpu <INT>          RAM to allocate to worker node(s) (Default: CAASP_WORKER_RAM=$WORKER_RAM)
+    --worker-cpu <INT>          CPUs to allocate to worker node(s) (Default: CAASP_WORKER_CPU=$WORKER_CPU)
+    --extra-repo <STR>          URL of a custom repository on the master(s)/worker(s) (Default: CAASP_EXTRA_REPO=$($DIR/../misc-tools/download-image --expand-channel --type kvm channel://devel))
```
kiall: Is the default not extra repos? This seems misleading :)

caasp-kvm/caasp-kvm
```diff
@@ -175,6 +182,13 @@ while [[ $# > 0 ]] ; do
       WORKER_CPU="$2"
       shift
       ;;
+    --extra-repo)
+      EXTRA_REPO="$2"
+      shift
+      ;;
+    --vanilla)
```
kiall: The `--vanilla)` section already exists above

MaximilianMeister:
@kiall updated

MaximilianMeister:
@mjura please have another look

PR #203, opened 2017-12-22
Update test cluster build instructions

stefsuse:
- now includes a step-by-step guide from fresh OS install to build

PR #211, opened 2018-01-05
KVM: Fix libvirt group name in readme

kiall:


stefsuse:
This change is also covered in #203

kiall:
No longer needed

PR #213, opened 2018-01-08
Testinfra: Allow for non machine-id hostnames

kiall:
As we are removing our machine-id hostnames, we'll need to support
both machine-id and unique-hostname for a short period of time. Add
support to check for either/or in tests that need it.

Once the machine-id hostnames patches have merged elsewhere, the
compatibility code here should be removed.


kiall:
Needed by https://github.com/kubic-project/salt/pull/340

PR #216, opened 2018-01-11
RFC: bring back caasp kvm qa project (simple erb version)

bs-suse:
The purpose of this PR is to fill the gap between automation/caasp-kvm and the original caasp-kvm project.
It is based on the discussion in PR #185 and the work @flavio has done as well as @kiall.

Without this PR automation/caasp-kvm creates caasp clusters where the admin node has development code injected. This is useful for development purposes, but it doesn't help when doing QA or debugging tasks.

With this PR the caasp-kvm script has been extended with a new flag: --vanilla. By default the flag is set to false, leading to the quick creation of "development" caasp clusters where the admin node does have the local development code injected.
On the opposite, when the --vanilla flag is used, the cluster will be created "vanilla", without any code injected to do qa testing or debugging.

Due to the lack of conditionals I resorted to the usage of an embedded ruby template to divert in the terraform description. This is then applied and evaluated by calling erb from the wrapper script that we use anyway. This has been done to reduce the code duplication.

I hope we can find consensus here and thereby close PR #185 along with this one.


caasp-kvm/caasp-kvm
```diff
@@ -8,6 +8,7 @@ DIR="$( cd "$( dirname "$0" )" && pwd )"
 ACTION=
 RUN_BUILD=
 RUN_DESTROY=
+export VANILLA=${VANILLA:-}
```
kiall: Rather than exporting the var, can we explicitly pass it though? Dropping the `export ` here, and using `VANILLA=$VANILLA erb T ...` below will make it explicit that that flag is being passed to erb for use,

(We had issues in our older tooling where it was very difficult to trace what vars were passed / used / available where, mostly due to relying on exported shell variables!)

bs-suse: For sure, we can. Done.

PR #224, opened 2018-01-19
Rotate workers automatically

FedericoCeratto:
On each run,
- offline workers that are not busy are deleted
- the online worker with longest runtime is put offline so that it will have time to drain and be deleted in one of the next runs.


jenkins-pipelines/Jenkinsfile.workers-rotation
```diff
@@ -0,0 +1,51 @@
...
+
+// Workers rotation
+
+import hudson.model.*
+
+// configuration
+def onlineWorkersThreshold = 3
+def offlineReason = "Disabled by worker rotation script"
+
+node {
```
kiall: node { should be unnecessary, there's nothing that will run on the node..

PR #229, opened 2018-01-25
Fast fail in CI

ereslibre:
If we get a failure or a success in CI make sure we do as early as possible,
without the need to explicitly wait the timeout when there's already a
final status (we only timeout if we are actually spinning).


ereslibre:
Rebased.

ereslibre:
Do not merge yet, it needs to be polished, it gives false positives sometimes.

velum-bootstrap/spec/features/02-bootstrap-cluster.rb
```diff
@@ -128,9 +128,17 @@
...
-        expect(page).to have_css(".fa-check-circle-o", count: node_number, wait: orchestration_timeout)
+        expect(page).to have_css(".fa-check-circle-o, .fa-times-circle", count: node_number, wait: orchestration_timeout)
       end
     end
     puts "<<< Orchestration completed"
+
+    puts ">>> Orchestration succeeded"
+    with_screenshot(name: :orchestration_succeeded) do
+      within(".nodes-container") do
+        expect(page).to have_css(".fa-check-circle-o", count: node_number)
```
vitoravelino: The default is set to 3 minutes, I would use `wait: 5` here.

ereslibre: Addressed 👍

PR #230, opened 2018-01-25
Fast fail in CI

ereslibre:
If we get a failure or a success in CI make sure we do as early as possible,
without the need to explicitly wait the timeout when there's already a
final status (we only timeout if we are actually spinning).

Backport of https://github.com/kubic-project/automation/pull/229


ereslibre:
Do not merge yet, it needs to be polished, it gives false positives sometimes.

PR #232, opened 2018-01-25
[Test CI] DO NOT MERGE - Run salt-minion with debug loglevel

ereslibre:


ereslibre:
This is to try to find the root cause of the only reason for flaky tests now: "Minion did not respond" error message.

PR #235, opened 2018-01-30
Fix image creation after some changes on bundler.

ereslibre:
Some symlinks were broken.


ereslibre:
This won't be required.

PR #246, opened 2018-02-07
use reverse dns names for minions

bs-suse:
This is important for DHCP assigned addresses in a bridged network.

### DIFF
--- a/caasp-kvm/tools/generate-environment
+++ b/caasp-kvm/tools/generate-environment
@@ -34,10 +34,17 @@ out=$(cat $TF_STATE | \
 for node in $(echo "$out" | jq -r '.minions[] | select(.["minion_id"]? == null) | [.addresses.publicIpv4] | join(" ")'); do
     machine_id=$(ssh root@$node $SSH_ARGS cat /etc/machine-id)
     out=$(echo "$out" | jq ".minions | map(if (.addresses.publicIpv4 == \"$node\") then . + {\"minionID\": \"$machine_id\"} else . end) | {minions: .}")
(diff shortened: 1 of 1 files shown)


caasp-kvm/tools/generate-environment
```diff
@@ -34,10 +34,21 @@ out=$(cat $TF_STATE | \
 for node in $(echo "$out" | jq -r '.minions[] | select(.["minion_id"]? == null) | [.addresses.publicIpv4] | join(" ")'); do
     machine_id=$(ssh root@$node $SSH_ARGS cat /etc/machine-id)
     out=$(echo "$out" | jq ".minions | map(if (.addresses.publicIpv4 == \"$node\") then . + {\"minionID\": \"$machine_id\"} else . end) | {minions: .}")
+    fqdn=$(echo $out | jq ".minions[] | select(.addresses.publicIpv4 == \"$node\") | .fqdn")
+    ip=$(dig +short $fqdn)
```
MaximilianMeister: for some weird reason, that i couldnt figure out yet, this doesnt return any result, so `$ip` is always empty here locally. only if i change it to `ip=$(echo $fqdn | xargs dig +short)` it works... any idea what that could be? it might be sth like bash implicit quoting, idk

bs-suse:
very well spotted @MaximilianMeister - fixed now

bs-suse:
I've rebased and had to do a push -f then. Not sure if there would have been a better way to do that, but the changes are the very same that @MaximilianMeister reviewed.

PR #247, opened 2018-02-07
suse registry mirror automation on openstack

ellisab:
Openstack automation for testing suse registry mirror feature for both http and https, depending if https is set to true or false


suse-registry-mirror-openstack-terraform/suse-registry-mirror-openstack-terraform
```diff
@@ -0,0 +1,87 @@
...
+
+if [ -v $OS_PASSWORD ]; then
+  echo "Please enter your OpenStack Password: "
+  read -sr OS_PASSWORD_INPUT
+  export OS_PASSWORD=$OS_PASSWORD_INPUT
+fi
+
+terraform $COMMAND -var auth_url=$OS_AUTH_URL -var domain_name=$OS_USER_DOMAIN_NAME \
+                    -var region_name=$OS_REGION_NAME -var project_name=$OS_PROJECT_NAME \
+                    -var user_name=$OS_USERNAME -var password=$OS_PASSWORD -var-file=openstack.tfvars
```
flavio: I think there's no need to pass all the `OS_` variables. They should already be extracted by the terraform openstack plugin

flavio:
I think you forgot to add the `.tf` file.

Another note: the setup scripts of the http and https registries are identical, they only different in the registry configuration file. I would avoid this duplication by either using a template for the file or by moving the registry configuration into dedicated files (one for http and another for https).

flavio:
The files look better now because there's no code duplication anymore. However I'm still skeptical about the explicit usage and definition of the openstack related variables. Are we sure they aren't set and looked up by default?

@mjura, @kiall and @inercia: could you take a quick look too?

kiall:
I also think the explict OS_* var processing is unnecessary - TF reads those by default, but it's harmless to be explicit!

PR #254, opened 2018-02-13
[WIP] Testinfra: Handle etcd role as a full role

kiall:


testinfra/tests/test_etcd.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def test_salt_role(self, host):
+        assert 'etcd' in host.salt("grains.get", "roles")
+
+    def test_salt_id(self, host):
+        machine_id = host.file('/etc/machine-id').content_string.rstrip()
+        assert machine_id in host.salt("grains.get", "id")
+
+    def test_etcd_aliveness(self, host):
+        hostname = host.file('/etc/hostname').content_string.rstrip()
```
dannysauer: If this comes out of "work in progress" mode, it might be worth switching this to either `os.uname()[1]` or `socket.gethostname()` instead of reading the file.  The `hostname` command uses a syscall to `uname`, so I'd normally be inclined to go with `os.uname` for this purpose despite the vague portability warning in the docs; this is unlikely to be run on HP-UX 11.11 or whatever.

https://docs.python.org/2/library/os.html#os.uname

MalloZup: @dannysauer good catch. :+1: 

I think we will need to figure out what we do with this PR since kiall is not anymore working at SUSE. 
We might backport it anyway with a new PR.

MalloZup:
@dannysauer  i am closing this for 3 reasons:

1) it is a minor refactoring from someone who doesn\t maintain the code anymore.

2) it add lot of bashism  ( this we should get rid in general :smile: )

3) there is no big values in this PR, we are not testing more a feature. closing

PR #255, opened 2018-02-14
Speed up/simplify pod scaling

FedericoCeratto:
Leaving the adaptive function around as it can be used for large number of pods

### DIFF
--- a/k8s-pod-tests/k8s-pod-tests
+++ b/k8s-pod-tests/k8s-pod-tests
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


k8s-pod-tests/k8s-pod-tests
```diff
@@ -165,13 +169,29 @@ scale_pod_in_steps() {
...
 
     [[ $current_num == $NUM_REPLICAS ]] && return
     sleep $cycle_time
 
   done
+}
+
+scale_pod_in_steps() {
+  # start from the currently *available* number
+  step=10
```
kiall: Can we add a `--step` param instead of hardcoding?

k8s-pod-tests/k8s-pod-tests
```diff
@@ -165,13 +169,29 @@ scale_pod_in_steps() {
...
+  current_num=$( get_deployments | awk '{print $5}' )
 
+  while [[ $current_num -lt $NUM_REPLICAS ]]; do
```
kiall: So - if we end up in a bad place, where the pods don't start for some reason, this look looks like it will hang forever. We probably need to add a `--step-timeout` flag with a reasonable default (180 seconds maybe?)

kiall:
Additionally --slowscale still requires an INTERVAL to be supplied, even though it's no longer used by that function. Maybe we should split off the INTERVAL and NUM_REPLICAS from both `--scale` and `--slowscale` into their own flags? This has the added bonus of being consistent with our other tools into the repo - flags take 1 arg each.

kiall:
This doesn't ever seem to complete now for some reason: http://jenkins.caasp.suse.net/blue/organizations/jenkins/automation/detail/PR-255/7/pipeline

PR #257, opened 2018-02-15
Velum Bootstrap: Expect HTTP 200 Status Codes

kiall:
If we don't do this, later checks such as this one will hang for the duration
of the wait period:

    expect(page).to have_selector( ..., wait: 400)


kiall:
@MaximilianMeister does that really simplify it? It would just swap 1 extra line next to every visit "/" for 2 and an indirection?

MaximilianMeister:
> does that really simplify it? It would just swap 1 extra line next to every visit "/" for 2 and an indirection?

i think so. it's a recurrent pattern, so i'd make it a function. otherwise you end up with a lot of DRY violations

the number of lines could be reduced by the other way of making a block statement:

```ruby
with_status_ok { visit "/a/page" }
```

PR #258, opened 2018-02-15
Add a 10 minute delay before accepting nodes

kiall:
This is (obviously) a total hack. We need to fix and fix the underlying issue - this should however unblock CI.


flavio:
This is an horrible hack, but if that's the only way to get the CI back... this would unblock us

kiall:
We've passed all the relevant parts of the CI for this for the first time in days - I'm going to merge this, despite it being *HORRIBLE*.

PR #264, opened 2018-02-22
Allow some extra time for Salt to recover after the admin node update

kiall:
The salt-master is taking a little more time than expected to recover
after the admin node upgrade reboot, lets give it a little extra time
to recover as a workaround.


kiall:
I accidently pushed this right into master.. derp. closing as it's "merged" now anyway

PR #272, opened 2018-02-27
Use sonobuoy e2e tests

ereslibre:


k8s-e2e-tests/entrypoint.sh
```diff
@@ -55,47 +55,13 @@ EOF
...
-
-    # For serial tests we have to set GINKGO_PARALLEL_NODES=1
-    export GINKGO_PARALLEL_NODES=1
-
-    # now run the serial tests
-    env E2E_REPORT_PREFIX=serial \
-    go run hack/e2e.go -v --test \
-        --test_args="$KUBE_TEST_ARGS --ginkgo.focus=\[Serial\].*\[Conformance\]" 2>&1 | \
-            tee -a /tmp/e2e-test.log
+    curl -L https://raw.githubusercontent.com/cncf/k8s-conformance/master/sonobuoy-conformance.yaml | kubectl apply -f -
```
kiall: I know it's WIP - but we should avoid pulling "master" so it doesn't break on us down the line.

kiall: Oh - actually - is it intended that you use master, and then it "does the right thing"? Humm

ereslibre: Yeah

ereslibre: Yes, it's intended. It will do the right thing based on the K8s version (1.8 or 1.9). However I took your comment as if something breaks upstream master and we are failing because of that. We could pin to a specific revision that we know it works, but I also wonder how often that will break since it's the conformance tests everyone will run... so I think it "should" be safe to use `master` directly? Also, this assumes github is not down at the moment of doing the http request... but in any case all our code is on GH and Jenkins still needs to fetch it...

ereslibre: As a side note: even if we pinned the specific revision of `k8s-conformance` it still uses external containers that we don't control, so even pinning the version of this repo does not ensure that we'll be breaking for whatever external reason.

ereslibre:
Didn't test it on jenkins, tested locally in separate steps.

ereslibre:
[Run new conformance tests on Jenkins](https://trello.com/c/Qs30JTN7/701-run-new-conformance-tests-on-jenkins)

PR #276, opened 2018-02-28
make download helper capable to handle SHA256 checksum files

bs-suse:
In addition to the .sha256 format (one file per checksum),
we now support the SHA256SUMS file format with all the checksums
for a directory.


kiall:
I don't understand why the CI job was killed - restarting it.

kiall:
It looks like this doesn't work for our usual use case - the get_sha256 method was missing a + '.sha256' in one branch, meaning python was downloading the entire qcow2.. Additionally, I needed to be explicit about passing `proxies` as a keyword argument to ensure it was used correctly.

    diff --git a/misc-tools/download-image b/misc-tools/download-image
    index 7aa9cca..1f2a4da 100755
    --- a/misc-tools/download-image
    +++ b/misc-tools/download-image
    @@ -121,12 +121,13 @@ def get_canonical_url(args, url):
             raise Exception("Cannot find image location")
     
     def get_sha256(url, proxies):
    +    print(" >> Downloading Sha256 File")
         partition = url.rpartition('/')
    -    sha256file = requests.head(url + '.sha256')
    +    sha256file = requests.head(url + '.sha256', proxies=proxies)
         
         if sha256file.status_code == 404:
             # try to find this file in /SHA256SUMS within the same directory
    -        for line in requests.get(partition[0] + "/SHA256SUMS", proxies).text.split('\n'):
    +        for line in requests.get(partition[0] + "/SHA256SUMS", proxies=proxies).text.split('\n'):
                 # find sha256 for this file
                 columns = line.split(" ")
                 if columns[-1] == partition[2]:
    @@ -135,10 +136,10 @@ def get_sha256(url, proxies):
             raise Exception("Cannot find remote image SHA256 in %s" % partition[0] + "/SHA256SUMS")
         elif sha256file.status_code == 302 or sha256file.status_code == 301:
             location = sha256file.headers.get('Location')
    -        return get_sha256(location[:location.rfind('.sha256')], proxies)
    +        return get_sha256(location[:location.rfind('.sha256')], proxies=proxies)
         elif sha256file.status_code == 200:
             # this will work if the SHA256 is stored as a PGP SIGNED MESSAGE in the .sha256 file
    -        sha256 = requests.get(url, proxies).text.split('\n')[3]
    +        sha256 = requests.get(url + '.sha256', proxies=proxies).text.split('\n')[3]
             if len(sha256) == 64:
                 return sha256
             raise Exception("Cannot find remote image SHA256 in %s" % url + '.sha256')

bs-suse:
I think what happens here is that the image (and the according SHA256 file) gets removed/replaced with newer ones during download. Therefore the .sha256 file is not there after the otherwise successful download.

So this can be considered as a failure we want to happen (as in: desired Exception raised) or as something we want to avoid/ignore.

Based on that, we have to decide if we even want to check the SHA after download (see https://github.com/bs-suse/automation/blob/2d95500401c7da83431ea0ef3b75786074bea61f/misc-tools/download-image#L174)

bs-suse:
I'll close this because I don't want it to live in my master. Moved to #286.

PR #280, opened 2018-03-05
Velum-Bootstrap: Add support for proxy and suse registry cache

kiall:


kiall:
This will be followed up by a caasp-kvm change to allow putting the necessary data in the env.json file

environment.json.example
```diff
@@ -46,5 +46,17 @@
...
+      "systemwide": true,
+      "http_proxy": "http://proxy.local:3128/",
+      "https_proxy": "http://proxy.local:3128/",
+      "no_proxy": ".local, localhost, 127.0.0.1"
+   },
+   "suse_registry_mirror": {
+      "enabled": true,
+      "url": "http://suse-mirror.local/",
+      "certificate": "FOO\nBAR\netc\n"
+  }
```
MaximilianMeister: one more intent >

velum-bootstrap/spec/support/helpers.rb
```diff
@@ -56,12 +56,52 @@ def configure
...
       check "settings[tiller]"
     else
       uncheck "settings[tiller]"
     end
+
+    # Proxy Settings Section
+    if environment.fetch("proxy", {}).fetch("enabled", false)
+      find('#settings_enable_proxy_enable').trigger('click')
+      # Nasty hack to let the panel open
+      sleep 2
```
MaximilianMeister: this is funny ;-) and ugly. maybe we could make it look nicer by adding a `with_panel_open` helper that accepts a selector and a &block? then it would be more consistent and you need to only comment it once

ereslibre: I wonder why we need this and how this can help to keep the panel open :/. I guess using `.click` instead of `.trigger('click')` is the same thing, right?

PR #287, opened 2018-03-15
Purify download file

bs-suse:
This change will make the use of local copies of remote URL images possible even if the remote image is gone.

It will make download_file stop trying to make the given url from args canonical. 
By that we avoid the http head request in make_canonical that would fail if there is no remote file (anymore). We would not even try to download the image anyway if a local copy of the same name is there. 
Beside that, making the url canonical is not needed anyway because we use wget to download and that tool takes care of redirects.

I hope that #286 will get merged, after that I'll rebase this to get it in.


kiall:
> I hope that #286 will get merged, after that I'll rebase this to get it in.

#286 is merged, can you rebase? Thanks!

PR #289, opened 2018-03-15
Fix terraform name prefix

bs-suse:
things I forgot last time


bs-suse:
the ci failure happened way later than the code that was touched in this PR is in the pipeline

PR #290, opened 2018-03-16
Add terraform bridged networking option

bs-suse:
This will give us tho option to use a bridged network interface and DHCP addresses for terraform deploys.


kiall:
I suspect this is using the bridged mode within CI, causing the error. I've pushed a change to the Jenkins config to archive the generated cluster.tf file to make debugging that kind of thing easier - https://github.com/kubic-project/jenkins-library/pull/148

kiall:
Yea, that change to archive cluster.tf merged - and this PR is indeed setting bridged mode in CI:

	  network_interface {
	 
	    bridge = ""
	    wait_for_lease = 1
	  }

bs-suse:
Mea culpa, I thought I tested without a bridge configured, but probably forgot to unset the environment variable and therefore tested two times with a bridge.

I've fixed the erb template now and tested both ways this time.

kiall:
Sorry - thought I merged this as I approved it last week

PR #291, opened 2018-03-16
Do not copy gems from the packaging side as they contain invalid symlinks

ereslibre:
In any case we are downloading gems from the outside (rubygems) because the
development gems are not included with the Velum package, so don't copy
gems back to development image.

Fixes the fontawesome problem.

### DIFF
--- a/caasp-kvm/tools/build-velum-image
+++ b/caasp-kvm/tools/build-velum-image
@@ -57,10 +57,10 @@ build_fresh_image() {
                     bash -c "
                       cd /srv/velum-latest && \
                       mkdir -p /var/lib/velum && \
-                      cp -r /srv/velum/vendor/bundle/ruby /var/lib/velum/ && \
-                      bundle config --local frozen 0 && \
-                      bundle config --local build.nokogiri --use-system-libraries && \
-                      bundle install --binstubs=/usr/local/bin --deployment --path=/var/lib/velum && \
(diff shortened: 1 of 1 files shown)


ereslibre:
[Velum: fix broken font-awesome](https://trello.com/c/95zU53VK/616-velum-fix-broken-font-awesome)

caasp-kvm/tools/build-velum-image
```diff
@@ -57,10 +57,10 @@ build_fresh_image() {
                     bash -c "
                       cd /srv/velum-latest && \
                       mkdir -p /var/lib/velum && \
-                      cp -r /srv/velum/vendor/bundle/ruby /var/lib/velum/ && \
-                      bundle config --local frozen 0 && \
-                      bundle config --local build.nokogiri --use-system-libraries && \
-                      bundle install --binstubs=/usr/local/bin --deployment --path=/var/lib/velum && \
+                      gem install --no-ri --no-rdoc bundler -n /usr/bin && \
+                      bundle.ruby2.1 config --local frozen 0 && \
```
MaximilianMeister: why is it `bundle.ruby2.1`? in my env (velum container) i only have `bundle`

ereslibre: This is because I'm installing `bundler` with `gem install bundler` and it doesn't have the good symlink. This is a huge hack and I'd prefer to have it solved in other way (more maintanable) but I think we can go over it at some other moment...

PR #292, opened 2018-03-20
When running bootstrap using caasp-devenv make sure we setup the local environment

ereslibre:
As a user, if I run `./caasp-devenv --build -m 3 -w 2 --bootstrap --testinfra` I
expect it to call to `velum-interactions --setup` too, so my local environment is
ready to run `rspec` locally.


ereslibre:
@MaximilianMeister you're right, missed `caasp-devenv --setup`. Closing this one.

PR #295, opened 2018-03-23
add pipeline to test packages from the sandbox repo

MaximilianMeister:

### depends on

https://github.com/kubic-project/jenkins-library/pull/146


MaximilianMeister:
[(5) devel sandbox](https://trello.com/c/7ZUWqVsF/707-5-devel-sandbox)

jenkins-pipelines/Jenkinsfile.kubic-sandbox
```diff
@@ -0,0 +1,28 @@
+def kubicLib = library("kubic-jenkins-library@${env.BRANCH_NAME}").com.suse.kubic
+
+// Configure the build properties
+properties([
+    buildDiscarder(logRotator(numToKeepStr: '31', daysToKeepStr: '31')),
+    disableConcurrentBuilds(),
```
kiall: We could add a paramater, to allow the extra repo to be changed if needed - e.g, just after `disableConcurrentBuilds`:

    parameters([
        string(name: 'EXTRA_REPO', defaultValue: 'devel:CaaSP:SandBox', description: 'Extra repo to add'),
    ])

then, the value can be used:

    kvmTypeOptions.extraRepo = env.EXTRA_REPO

this would make it a little more flexible.

jenkins-pipelines/Jenkinsfile.kubic-sandbox
```diff
@@ -0,0 +1,28 @@
...
+])
+
+def kvmTypeOptions = kubicLib.CaaspKvmTypeOptions.new();
+kvmTypeOptions.extraRepo = 'devel:CaaSP:SandBox'
+
+coreKubicProjectPeriodic(
+    environmentTypeOptions: kvmTypeOptions
+) {
+    // Run through the upgrade orchestration
+    upgradeEnvironment(
```
kiall: So, I think this job will only use the new packages at this point - right?

jenkins-pipelines/Jenkinsfile.kubic-sandbox
MaximilianMeister: i'll re-add it thx (already had it in but removed again)

jenkins-pipelines/Jenkinsfile.kubic-sandbox
MaximilianMeister: good question, if there's not the same package in the testupdate repo i assume... but i am not fully sure

jenkins-pipelines/Jenkinsfile.kubic-sandbox
MaximilianMeister: added

MaximilianMeister:
@kiall i think we can close this, as we're better off using images for that, as you suggested. not sure if the code here is still beneficial in any way

PR #298, opened 2018-03-26
enable adding/removing nodes through caasp-kvm and velum-interactions

MaximilianMeister:

## how to test

```
# spawn and bootstrap cluster
./caasp-devenv --build -m 1 -w 1 --bootstrap
# let terraform add another node
./caasp-devenv --update-deployment -m 1 -w 2
# run ui automation to add the node
./velum-interactions --node-add
# run ui automation to remove a node
./velum-interactions --node-remove
```

### nightly pipelines depend on:

https://github.com/kubic-project/jenkins-library/pull/151

### DIFF
--- a/jenkins-pipelines/Jenkinsfile.kubic-nightly-node-add
+++ b/jenkins-pipelines/Jenkinsfile.kubic-nightly-node-add
@@ -1,0 +1,34 @@
+def kubicLib = library("kubic-jenkins-library@${env.BRANCH_NAME}").com.suse.kubic
+
+// Configure the build properties
+properties([
+    buildDiscarder(logRotator(numToKeepStr: '31', daysToKeepStr: '31')),
+    disableConcurrentBuilds(),
+    pipelineTriggers([cron('@daily')])
+])
+
+def kvmTypeOptions = kubicLib.CaaspKvmTypeOptions.new();
+kvmTypeOptions.vanilla = true
+
+coreKubicProjectPeriodic(
+    environmentTypeOptions: kvmTypeOptions
+) {
+    // empty preBootstrapBody
+} {
+    // create another worker node
+    updateEnvironmentCaaspKvm(
+        environment: environment,
+        masterCount: 3,
+        workerCount: 3
+    )
+    // register it via velum
+    bootstrapNewNode(
+        environment: environment
+    )
+} {
+    // Run the Core Project Tests again
+    coreKubicProjectTests(
+        environment: environment,
+        podName: 'default'
+    )
+}
--- a/velum-bootstrap/spec/features/02-bootstrap-cluster.rb
+++ b/velum-bootstrap/spec/features/02-bootstrap-cluster.rb
@@ -1,5 +1,5 @@
 require "spec_helper"
-require 'yaml'
+require "yaml"
 
 feature "Boostrap cluster" do
 
@@ -144,7 +144,7 @@ feature "Boostrap cluster" do
     end
     puts "<<< Orchestration completed"
 
-    puts ">>> Orchestration succeeded"
+    puts ">>> Checking if orchestration succeeded"
     with_screenshot(name: :orchestration_succeeded) do
       within(".nodes-container") do
         expect(page).to have_css(".fa-check-circle-o", count: node_number, wait: 5)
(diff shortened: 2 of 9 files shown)


MaximilianMeister:
[Add functional tests to test add/removal of nodes](https://trello.com/c/qgaJQlrX/237-add-functional-tests-to-test-add-removal-of-nodes)

velum-bootstrap/velum-interactions
```diff
@@ -95,6 +101,9 @@ while [[ $# > 0 ]] ; do
       HAS_INTERACTION=true
       HAS_ACTION=true
       ;;
+    --node-add)
+      RUN_NODE_ADD=true
```
kiall: Would also need to have these two so it can be ran by itself:

       HAS_INTERACTION=true
       HAS_ACTION=true

velum-bootstrap/spec/features/02-bootstrap-cluster.rb
```diff
@@ -44,9 +49,22 @@
     end
     puts "<<< accept-all button enabled"
 
-    puts ">>> Click to accept all minion keys"
-    with_screenshot(name: :accept_button_click) do
-      click_button("accept-all")
+    # leave one minion for later acceptance to test the add node feature
+    if node_add_enabled?
```
kiall: I'm not sure this will work as expected. For example, testinfra will expect all nodes in env.json to be built+ready when it runs. I'd rather see us set a node in environment.json to an empty list of `roles: []` - indicating "don't use" for bootstrap (everything should handle this already), we just need a way to tell caasp-kvm it should build some extra nodes, but not assign a role to them - then when we later call `velum-interactions --add-node` - it would pick one of the unassigned nodes, run the 06-add spec, and finally set the new role in env.json for the other tools to pick up on later.

velum-bootstrap/spec/features/06-node-add.rb
```diff
@@ -0,0 +1,105 @@
+require "spec_helper"
+require "yaml"
+
+feature "Add a Node" do
+
+  let(:new_nodes) { environment["minions"].collect { |element| element["role"].empty? } }
```
MaximilianMeister: forgot to remove it, it's a bit misleading and it's not used

velum-bootstrap/spec/support/helpers.rb
```diff
@@ -57,7 +57,7 @@ def configure
     end
 
     fill_in "settings_dashboard", with: environment["dashboardHost"] || default_ip_address
-    if ENV.fetch("ENABLE_TILLER", false)
+    if ENV.fetch("ENABLE_TILLER", false) && ENV.fetch("ENABLE_TILLER") == "true"
```
ereslibre: This confuses me a bit, why not just replace `ENV.fetch("ENABLE_TILLER", false)` with `ENV.fetch("ENABLE_TILLER", false) == "true"`?

MaximilianMeister: i fixed it this way because the default is `ENABLE_TILLER=false` - but i agree it can be done shorter in one statement

jenkins-pipelines/Jenkinsfile.kubic-nightly-node-add
```diff
@@ -0,0 +1,34 @@
...
+properties([
+    buildDiscarder(logRotator(numToKeepStr: '31', daysToKeepStr: '31')),
+    disableConcurrentBuilds(),
+    pipelineTriggers([cron('@daily')])
+])
+
+def kvmTypeOptions = kubicLib.CaaspKvmTypeOptions.new();
+kvmTypeOptions.vanilla = true
+
+coreKubicProjectPeriodic(
```
kiall: Syntax is these files is a little off, will fixup after it merges (they won't activate straight away anyway)

velum-bootstrap/spec/features/07-node-remove.rb
```diff
@@ -0,0 +1,54 @@
...
+  append_after(:each) do
+    Capybara.reset_sessions!
+  end
+
+  scenario "User removes a node" do
+    with_status_ok do
+      visit "/"
+    end
+
+    puts ">>> Checking if node can be removed"
```
kiall: I'd love to be more explicit here - allowing for selecting a specific node to be removed rather than selecting "at random".

PR #312, opened 2018-04-09
master nodes are removable now as well

MaximilianMeister:
5 node nightly ci failed because it found 5 nodes to remove

from http://jenkins.caasp.suse.net/job/caasp-nightly-node-remove/job/master/7/consoleFull

```
>>> Checking if node can be removed
  User removes a node (FAILED - 1)

Failures:

  1) Remove a Node User removes a node
     Failure/Error: expect(page).to have_link(text: "Remove", count: worker_number, wait: 120)
       expected to find link nil with text "Remove" 2 times, found 5 matches: "Remove", "Remove", "Remove", "Remove", "Remove". Also found "kubectl config", "", "", which matched the selector but not all filters.
```

**cc** @vitoravelino can you confirm that masters are also removable now?

### DIFF
--- a/velum-bootstrap/spec/features/07-node-remove.rb
+++ b/velum-bootstrap/spec/features/07-node-remove.rb
@@ -4,7 +4,6 @@ require "yaml"
 feature "Remove a Node" do
 
   let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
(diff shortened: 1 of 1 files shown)


vitoravelino:
That's correct. One check that we might want to do is: there's no remove link when there's only 1 master or 1 worker. So I guess it's worth to check that before checking the number of remove.

PR #313, opened 2018-04-09
expect a modal when recommended cluster topology is broken

MaximilianMeister:

### DIFF
--- a/velum-bootstrap/spec/features/07-node-remove.rb
+++ b/velum-bootstrap/spec/features/07-node-remove.rb
@@ -33,11 +33,8 @@ feature "Remove a Node" do
       first(".remove-node-link").click
     end
 
-    # if we are going to remove a master node in the future explicitly
-    # remember to check if the number of masters are even
-    if (node_number-1) < 3
+    if page.has_content?("Unsupported cluster topology", wait: 5)
       with_screenshot(name: :unsupported_topology_modal) do
-        expect(page).to have_content("Unsupported cluster topology")
         click_button "Proceed anyway"
       end
     end


velum-bootstrap/spec/features/07-node-remove.rb
```diff
@@ -33,9 +34,7 @@
       first(".remove-node-link").click
     end
 
-    # if we are going to remove a master node in the future explicitly
-    # remember to check if the number of masters are even
-    if (node_number-1) < 3
+    if master_number.odd? || (node_number-1) < 3
```
kiall: I'm not sure this is right either? Since the node to be removed is selected at random (the first node, sorted by the machine UUID), theres no guarantee the first node is going to be a master.

vitoravelino: There's a scenario that might fail. If the removed node is a worker and it's a fine removal, modal won't appear and it will fail. We need to know if the node to be removed is a master or not when checking that.

vitoravelino: I didn't reload and see @kiall comment xD

MaximilianMeister: @vitoravelino can we then just check the page for the modal after all?

vitoravelino: Well, from the Capybara doc [0], `has_text?`/`has_content?` waits the default timeout. So I guess it's safe to go with that approach of checking if there's the text and then clicking on `Proceed anyway`. I didn't know that... We should try it again then.

[0] http://www.rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Matchers%3Ahas_content%3F

MaximilianMeister: yes, my argument was that the default wait time would be wasted if we know that there is no modal coming. but i guess it doesnt matter too much, i think 5 seconds is good enough

kiall:
This looks better - avoiding the duplicated logic seems like a good idea to me :)

MaximilianMeister:
[Add functional tests to test add/removal of nodes](https://trello.com/c/qgaJQlrX/237-add-functional-tests-to-test-add-removal-of-nodes)

PR #314, opened 2018-04-09
Add Netdata installer and metrics collector

FedericoCeratto:
Install Netdata on admin or master nodes for testing and debugging.

### DIFF
--- a/misc-tools/netdata/capture/README.md
+++ b/misc-tools/netdata/capture/README.md
@@ -1,0 +1,5 @@
+## Netdata charts capture tool
+
+Collect charts and badges
+
+Usage: see ./capture-charts --help
--- a/misc-tools/netdata/capture/capture-charts
+++ b/misc-tools/netdata/capture/capture-charts
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+type python3 >/dev/null 2>&1 || sudo zypper install --no-confirm python3
+
+if [ ! -d ~/py3venv ]; then
+  python3 -m venv ~/py3venv
+fi
+~/py3venv/bin/pip3 install requests pygal
+~/py3venv/bin/python3 $(dirname $0)/capture.py $@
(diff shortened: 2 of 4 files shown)


misc-tools/install-netdata
```diff
@@ -0,0 +1,36 @@
...
+import json
+import subprocess
+
+def runcmd(cmd):
+    print(cmd)
+    subprocess.check_call(cmd, shell=True)
+
+def main():
+    ap = argparse.ArgumentParser(description='Install Netdata on host')
+    ap.add_argument('target', choices=["admin", "master"], help="Target host")
```
kiall: Any reason to not include the workers?

misc-tools/install-netdata
```diff
@@ -0,0 +1,36 @@
...
+    ap.add_argument('--env-json-path', default='./environment.json', help="environment.json full path")
+    args = ap.parse_args()
+
+    with open(args.env_json_path) as f:
+        env = json.load(f)
+    sshuser = env['sshUser']
+    sshkey = env['sshKey']
+    target = [m for m in env["minions"] if m["role"] == args.target][0]
+    ipaddr = target["addresses"]["privateIpv4"]
+    basecmd = "/usr/bin/ssh -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -i {} {}@{} ".format(sshkey, sshuser, ipaddr)
```
kiall: This can be simplified to `ssh -F environment.ssh_config %s".format(${minion[fqdn]})` with the added advantage of working with the proxycommand setup used on OpenStack etc.

kiall: (Though, maybe we want to have the environment.ssh_config path baked into environment.json like we reference ssh keys or kubeconfigs to make finding it easier)

FedericoCeratto:
Related to https://github.com/kubic-project/jenkins-library/pull/158
Output example: http://jenkins.caasp.suse.net/job/jenkins-library/job/PR-158/11/artifact/netdata/admin/charts/charts.html

lcavajani:
I agree with @kiall, we should add the workers too. Aside from that, I've tested it and it's great.

PR #322, opened 2018-04-12
correctly check for failed node removal

MaximilianMeister:

### DIFF
--- a/velum-bootstrap/spec/features/07-node-remove.rb
+++ b/velum-bootstrap/spec/features/07-node-remove.rb
@@ -45,20 +45,29 @@ feature "Remove a Node" do
       end
     end
 
-    orchestration_timeout = [[3600, 120].max, 7200].min
-    puts ">>> Waiting for node removal"
-    with_screenshot(name: :wait_node_removal) do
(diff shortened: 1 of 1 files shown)


velum-bootstrap/spec/features/07-node-remove.rb
```diff
@@ -45,20 +45,29 @@
...
+    end
+    puts "<<< node removal pending"
+
     orchestration_timeout = [[3600, 120].max, 7200].min
-    puts ">>> Waiting for node removal"
+    puts ">>> Waiting for node removal to be done"
     with_screenshot(name: :wait_node_removal) do
       within(".nodes-container") do
-        expect(page).to have_css(".fa-check-circle-o, .fa-times-circle", count: 1, wait: orchestration_timeout)
+        expect(page).not_to have_text("Pending removal", count: 1, wait: orchestration_timeout)
```
kiall: Does the `count: 1` make sense here?

MaximilianMeister: good question, probably count 1 is always the default anyways, as capybara only complains when there are more than 1 selected elements found... i'll drop it

PR #327, opened 2018-04-16
fix env check

MaximilianMeister:

### DIFF
--- a/jenkins-pipelines/Jenkinsfile.kubic-sandbox
+++ b/jenkins-pipelines/Jenkinsfile.kubic-sandbox
@@ -34,7 +34,7 @@ coreKubicProjectPeriodic(
         podName: 'default'
     )
 
-    if (env.RUN_CONFORMANCE_TESTS) {
+    if (env.RUN_CONFORMANCE_TESTS == "true") {
         stage('Run K8S Conformance Tests') {
             runTestK8sConformance(environment: environment)
         }


jenkins-pipelines/Jenkinsfile.kubic-sandbox
```diff
@@ -34,7 +34,7 @@ coreKubicProjectPeriodic(
         podName: 'default'
     )
 
-    if (env.RUN_CONFORMANCE_TESTS) {
+    if (RUN_CONFORMANCE_TESTS) {
```
kiall: Testing this out with replay, it looks like Jenkins treats all arguments as strings - the "booleanParam" just indicates which UI will be shown.

We'll need:

    if (RUN_CONFORMANCE_TESTS == "true") {

kiall: http://jenkins.caasp.suse.net/view/Misc%20Jobs/job/caasp-manual-sandbox/job/master/7 is rebuilding with this to verify:

    if (RUN_CONFORMANCE_TESTS == "true") {

kiall: `env.FOO` will work, but the variable starts in jenkins as just `FOO` and can be accessed either way... But `env.FOO` works - so merging :)

MaximilianMeister: i see, i was only wondering why http://jenkins.caasp.suse.net/view/Misc%20Jobs/job/caasp-manual-sandbox/job/master/7 didnt run the conformance tests... but i guess you've started it with default parameter (RUN_CONFORMANCE_TESTS=false)

PR #329, opened 2018-04-19
add a kubic channel

MaximilianMeister:

### DIFF
--- a/caasp-kvm/caasp-kvm
+++ b/caasp-kvm/caasp-kvm
@@ -12,7 +12,7 @@ RUN_UPDATE_DEPLOYMENT=
 RUN_PLAN=
 
 # caasp or kubic
-DIST=${DIST:-caasp}
+DIST=${CAASP_DIST:-caasp}
 
 MASTERS=${CAASP_NUM_MASTERS:-1}
 WORKERS=${CAASP_NUM_WORKERS:-2}
@@ -278,9 +278,6 @@ fi
 if [ -n "$CAASP_NAME_PREFIX" ] ; then
   TF_ARGS="$TF_ARGS -var name_prefix=$CAASP_NAME_PREFIX"
   log "Using name prefix for terraform: $CAASP_NAME_PREFIX"
-else
-  TF_ARGS="$TF_ARGS -var name_prefix=${DIST}_"
-  log "Using name prefix for terraform: $DIST"
 fi
 
 # Core methods
--- a/caasp-kvm/caasp-kvm
+++ b/caasp-kvm/caasp-kvm
@@ -11,6 +11,9 @@ RUN_DESTROY=
 RUN_UPDATE_DEPLOYMENT=
 RUN_PLAN=
 
+# caasp or kubic
+DIST=${DIST:-caasp}
+
 MASTERS=${CAASP_NUM_MASTERS:-1}
 WORKERS=${CAASP_NUM_WORKERS:-2}
 IMAGE=${CAASP_IMAGE:-channel://devel}
@@ -87,6 +90,10 @@ Usage:
   Build a 1 master, 2 worker cluster
 
   $0 --build -m 1 -w 2
+
+  Build a 1 master, 2 worker cluster using the latest kubic image
+
+  $0 --build -m 1 -w 2 --image channel://kubic
 
   Build a 1 master, 2 worker cluster using the latest staging A image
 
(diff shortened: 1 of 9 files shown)


sysrich:
I do not have write access so my review is not sufficient for this PR :)

caasp-kvm/cluster.tf.erb
```diff
@@ -22,74 +23,74 @@ variable "pool" {
 # Cluster variables #
 #####################
 
-variable "caasp_img_source_url" {
+variable "<%= dist %>_img_source_url" {
```
kiall: For the variables, I think it makes sense to just remove the caasp/kubic prefix alltogether

caasp-kvm/cluster.tf.erb
```diff
@@ -115,9 +116,9 @@ provider "libvirt" {
 }
 
 # This is the CaaSP kvm image that has been created by IBS
-resource "libvirt_volume" "caasp_img" {
-  name   = "${var.name_prefix}${basename(var.caasp_img_source_url)}"
-  source = "../downloads/kvm-${basename(var.caasp_img_source_url)}"
+resource "libvirt_volume" "<%= dist %>_img" {
```
kiall: Rather than prefix everything with dist - what about prefixing everything with "${var.name_prefix}" - and setting the prefix on the outside to be caasp/kubic by default..

caasp-kvm/cluster.tf.erb
```diff
@@ -126,33 +127,33 @@ resource "libvirt_volume" "caasp_img" {
 ##############
 <% if ENV['CAASP_BRIDGE'] == '' -%>
 resource "libvirt_network" "network" {
-    name      = "${var.name_prefix}caasp-net"
-    mode      = "${var.caasp_net_mode}"
-    domain    = "${var.name_prefix}${var.caasp_domain_name}"
-    addresses = ["${var.caasp_net_network}"]
+    name      = "${var.name_prefix}<%= dist %>-net"
```
kiall: If we keep the name_prefix, and the dist - ideally we wouldn't use two different templating languages at once - ideally we'd keep it as "terraform native" as possible :)

caasp-kvm/tools/build-velum-image
```diff
@@ -58,9 +66,9 @@ build_fresh_image() {
                       cd /srv/velum-latest && \
                       mkdir -p /var/lib/velum && \
                       gem install --no-ri --no-rdoc bundler -n /usr/bin && \
-                      bundle.ruby2.1 config --local frozen 0 && \
-                      bundle.ruby2.1 config --local build.nokogiri --use-system-libraries && \
-                      bundle.ruby2.1 install --deployment --binstubs=/usr/local/bin --path=/var/lib/velum && \
+                      bundle.${RUBY_VERSION} config --local frozen 0 && \
```
kiall: Can we instead fix the images, so these suffixes aren't needed? These are brittle anyway and will break whenever versions change (Happy to leave this as is for now - fixing the images is clearly a different issue)

caasp-kvm/cluster.tf.erb
MaximilianMeister: :smile: it's funny, i tried that and this is how it looked like:

`"${var.${var.dist}_net_mode}"`

caasp-kvm/cluster.tf.erb
MaximilianMeister: good idea :+1:

caasp-kvm/cluster.tf.erb
```diff
@@ -126,33 +126,33 @@ resource "libvirt_volume" "caasp_img" {
...
+  name      = "${var.name_prefix}admin_cloud_init.iso"
   pool      = "${var.pool}"
   user_data = "${file("cloud-init/admin.cfg")}"
 }
 
 resource "libvirt_domain" "admin" {
-  name      = "${var.name_prefix}caasp_admin"
-  memory    = "${var.caasp_admin_memory}"
-  vcpu      = "${var.caasp_admin_vcpu}"
-  metadata   = "${var.name_prefix}caasp-admin.${var.caasp_domain_name},admin,${count.index}"
```
kiall: Looks like CI is failing due to the change on this line.

Original:

    ${var.name_prefix}caasp-admin.${var.domain_name}

Original Produces:

    caasp-admin.devenv.caasp.suse.net

Updated:

    ${var.name_prefix}admin.${var.domain_name}

Updated Produces:

    caasp_admin.devenv.caasp.suse.net

The prefix here (and for masters/workers) really isn't needed, lets, remove it from the DNS name (i.e. the metadata field) so we end up with "admin.devenv.caasp.suse.net" etc - I've added these to the DNS server.

caasp-kvm/caasp-kvm
```diff
@@ -10,6 +10,9 @@ RUN_BUILD=
 RUN_DESTROY=
 RUN_UPDATE_DEPLOYMENT=
 
+# caasp or kubic
+DIST=${DIST:-caasp}
```
kiall: Nit: This should be:

    DIST=${CAASP_DIST:-caasp}

Yes, this has caasp in it. That's annoying, but it matches the pattern used for all other variables and prevents the scripts pulling in a preexisting DIST env var, and allows for easily identify what env vars are affecting an environment etc.. At some point, we might want to find another prefix for these vars...

MaximilianMeister: makes sense. i think caasp is ok for now, it's only internal and it's consistent

caasp-kvm/tools/build-velum-image
```diff
@@ -12,11 +12,19 @@ VELUM_DEVEL_IMAGE="$DIR/../resources/docker-images/velum-development.tar"
...
+  IMAGE_DIST=kubic
+  RUBY_VERSION=ruby2.5
+else
+  IMAGE_DIST=sles12
+  RUBY_VERSION=ruby2.1
+fi
 # Image reference
 if [[ $IMAGE == channel://* ]]; then
-  IMAGE_REFERENCE=$(jq -r ".baseurl.${IMAGE:10}" $REGISTRY_URLS_JSON)/sles12/velum-development:0.0
+  IMAGE_REFERENCE=$(jq -r ".baseurl.${IMAGE:10}" $REGISTRY_URLS_JSON)/$IMAGE_DIST/velum-development:0.0 # FIXME path to tag
```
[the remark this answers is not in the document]
MaximilianMeister: good idea! i had to adapt it a little bit, but this was very useful

PR #332, opened 2018-04-24
fix suse registry urls: paths do not contain arch

jordimassaguerpla:


jordimassaguerpla:
possible. Let me fire a retrigger build.

PR #338, opened 2018-04-25
[2.1] Pin to a 1.8 compatible sonobuoy

kiall:


mjura:
Just wondering because right now we rely on image `gcr.io/heptio-images/kube-conformance:v1.9` what if they will decide to remove it. Should we fork or branched this image somehow, somewhere?

kiall:
> Just wondering because right now we rely on image gcr.io/heptio-images/kube-conformance:v1.9 what if they will decide to remove it. Should we fork or branched this image somehow, somewhere?

I think that's less likely, but, worth considering. Maybe we should open an issue upstream and ask for branches / tags where older versions, even if unsupported, can pull from?

PR #339, opened 2018-04-25
[2.0] Pin to a 1.8 compatible sonobuoy

kiall:


mjura:
Just wondering because right now we rely on image gcr.io/heptio-images/kube-conformance:v1.9 what if they will decide to remove it. Should we fork or branched this image somehow, somewhere?

PR #343, opened 2018-04-25
fix orchestration check logic

MaximilianMeister:


velum-bootstrap/spec/features/06-node-add.rb
```diff
@@ -84,20 +84,15 @@
...
-      end
-    end
-    puts "<<< Node add orchestration completed"
-
-    puts ">>> Checking if node add orchestration succeeded"
-    with_screenshot(name: :node_add_orchestration_succeeded) do
-      within(".nodes-container") do
-        expect(page).to have_css(".fa-check-circle-o", count: new_nodes_count, wait: 5)
+        using_wait_time orchestration_timeout do
+          expect(page).not_to have_css(".fa-times-circle", count: new_nodes_count)
```
vitoravelino: When a node is added, there will be no failed state (only success and pending) and this expectation will pass because the condition is satisfied. You can either check for `not having pending (fa-spin)` or do the same as we did on bootstrap `success + fail == new total`. 

Keep in mind that `not_to` is tricky. I personally try to avoid it as much as possible.

velum-bootstrap/spec/features/06-node-add.rb
```diff
@@ -87,15 +89,15 @@
     puts ">>> Wait until node add orchestration is complete (Timeout: #{orchestration_timeout})"
     with_screenshot(name: :node_add_orchestration_complete) do
       within(".nodes-container") do
-        expect(page).to have_css(".fa-check-circle-o, .fa-times-circle", count: new_nodes_count, wait: orchestration_timeout)
+        expect(page).to have_css(".fa-check-circle-o, .fa-times-circle", count: node_number, wait: orchestration_timeout)
```
MaximilianMeister: @vitoravelino tested this locally now, all nodes are green after orchestrating the new node, but it keeps stuck in this loop :-(

PR #357, opened 2018-04-30
only by --setup option clone  fondamental github repository for kubic

MalloZup:
i had some difficulties to clone all needed repos needed by other scripts.

I think having an array of github_repo can improve to manage/maintain also this repos in subprojects at least for the casp-kvm-part which is relevant for the casp-dev-env

### DIFF
--- a/caasp-devenv
+++ b/caasp-devenv
@@ -54,7 +54,11 @@ DIST_PACKAGES="jq \
                docker \
                libvirt-daemon-qemu \
                guestfs-tools \
+               git \
                ca-certificates-suse"
+
+# kubic-project needed repos to clone via https on build(sep by 1 space)
+KUBIC_NEEDED_GITHUB_REPOS="caasp-container-manifests velum salt"
 
 USAGE=$(cat <<USAGE
 Usage:
@@ -127,6 +131,19 @@ warn()       { log "WARNING: $@" ; }
 error()      { log "ERROR: $@" ; exit 1 ; }
 check_file() { [ -f "$1" ] || error "File $1 doesn't exist!" ; }
 usage()      { echo "$USAGE" ; exit 0 ; }
+
+# clone needed github repos (dependencies)
+clone_needed_github_repos() {
+  git_repos="$1"
+  pushd $DIR/../
+  for git_repo in $git_repos; do
+  	if [ ! -d "$git_repo" ]; then
+             git clone "https://github.com/kubic-project/$git_repo.git"
+        fi
+  done
+  popd
+} 
+
 
 # parse options
 while [[ $# > 0 ]] ; do
@@ -236,6 +253,9 @@ setup() {
 
   # Ensure the current user has access to docker and libvirt
   sudo usermod -aG docker,libvirt $USER
+
+  # clone needed kubic github repos
+  clone_needed_github_repos "$KUBIC_NEEDED_GITHUB_REPOS"
 
   (cd "$DIR/velum-bootstrap" && ./velum-interactions --setup)
 


caasp-devenv
```diff
@@ -128,6 +132,24 @@ error()      { log "ERROR: $@" ; exit 1 ; }
 check_file() { [ -f "$1" ] || error "File $1 doesn't exist!" ; }
 usage()      { echo "$USAGE" ; exit 0 ; }
 
+# clone needed github repos (dependencies)
+clone_needed_github_repos() {
+  git_repos="$1"
+  current=`pwd`
```
MalloZup: maybe some bashism magic can improve the cd logic :jack_o_lantern:

caasp-devenv
```diff
@@ -128,6 +132,24 @@ error()      { log "ERROR: $@" ; exit 1 ; }
...
+  cd ..
+  for git_repo in $git_repos; do
+  	if [ ! -d "$git_repo" ]; then
+             git clone "https://github.com/kubic-project/$git_repo.git"
+        else
+             cd $git_repo
+             git pull
```
stefsuse: If a git pull is unsuccessful this should probably exit.

caasp-devenv
```diff
@@ -237,6 +259,9 @@ setup() {
   # Ensure the current user has access to docker and libvirt
   sudo usermod -aG docker,libvirt $USER
 
+  # clone needed kubic github repos
+  clone_needed_github_repos "$KUBIC_NEEDED_GITHUB_REPOS"
```
stefsuse: This will only get called on setup - can you move this so it also updates on --build?

caasp-devenv
kiall: Any command that fails in this script will cause it to exit straight away - the `set -euo pipefail` at the top ensures this (specifically, the `set -e` part of it)

caasp-devenv
kiall: This won't always end up with the path you expect - e.g. try `./automation/caasp-devenv ...` - the repos will end up cloned an extra folder up. The `DIR` variable at the top works around this...

However - It would be better to just do this:

    pushd $DIR/../ <-- brings you to the same folder as `automation` is cloned to.
    git clone ....
    git clone ....
    popd <-- returns you to wherever you came from..

caasp-devenv
kiall: A slightly better option than a plain `git pull` would be a `git pull --rebase` - which means any local/unpublished commits will be dropped on top of the current remote HEAD without creating a merge commit.

caasp-devenv
kiall: IMO, that's not a good idea... When developing on Kubic/CaaSP, you don't want your clones to change just because you're rebuilding the virtual machines.. Maybe adding a `--git-pull` flag would cover the use case?

That way, if you don't care about any current repo state, you can `./caasp-devenv --git-pull --build --bootstrap` and get the latest code in a fresh environment?

MalloZup:
Thx for reviews, i will adress soon the comment

caasp-devenv
MalloZup: thx i was searching for this :ice_skate:

MalloZup:
i removed updating github repository because it can bring extra failures, eg

```
git clone kubic
git remote add myfork
git  branch newbranchy
git checkout newbranchy 
git pull --rebase

```

caasp-devenv
MalloZup: from my pov we are fine to have this only on build

PR #358, opened 2018-05-02
Schedule all nightly jobs between 3 and 5am UTC

kiall:
This will help ensure periodic jobs are running when the CI system
is otherwise idle outside of EU / US working hours.


FedericoCeratto:
Perhaps we could start each task at a fixed time to make the whole daily run more deterministic. It also makes it easier to spot automated runs from manual runs in the history.

PR #366, opened 2018-05-03
enable rubocop for rspec velum-bootstrap

MalloZup:
- [x] enable rubocop for ```velum-bootsrap``` repo.
- [x] fix small issues
- [x] wait for ci results on pr
- [x] tell to  activate travis for repo automation
- [x] activate rubocop on forked repo


velum-bootstrap/spec/features/02-bootstrap-cluster.rb
```diff
@@ -1,63 +1,60 @@
...
-  let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
-  let(:hostnames) { environment["minions"].map { |m| m["fqdn"] if m["role"] != "admin" }.compact }
-  let(:minion_ids) { environment["minions"].map { |m| m["minionId"] if m["role"] != "admin" }.compact }
-
-  before(:each) do
-    unless self.inspect.include? "User registers"
-      login
-    end
+  before do
+    login unless inspect.include? 'User registers'
```
MalloZup: self is rendundant

velum-bootstrap/spec/features/02-bootstrap-cluster.rb
```diff
@@ -1,63 +1,60 @@
...
 
-  let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
-  let(:hostnames) { environment["minions"].map { |m| m["fqdn"] if m["role"] != "admin" }.compact }
-  let(:minion_ids) { environment["minions"].map { |m| m["minionId"] if m["role"] != "admin" }.compact }
-
-  before(:each) do
-    unless self.inspect.include? "User registers"
-      login
-    end
+  before do
```
MalloZup: each in before is default

MalloZup:
@MaximilianMeister @kiall  i didn't fixed all rubocop lint-error but created with

```rubocop --auto-gen-config````whitelist for this.

For this pr we need to activate travis-ci on this PR.

**Comment by MalloZup** [comment]
you can see here travis https://github.com/MalloZup/automation/pull/1

File: velum-bootstrap/.rubocop_todo.yml [hunk 3]
```diff
@@ -0,0 +1,85 @@
+# This configuration was generated by
```
MaximilianMeister: can we use most of https://github.com/kubic-project/velum/blob/master/.rubocop.yml here? that keeps both projects consistent

velum-bootstrap/.rubocop_todo.yml
```diff
@@ -0,0 +1,85 @@
...
+# `rubocop --auto-gen-config`
+# on 2018-05-03 15:23:52 +0200 using RuboCop version 0.55.0.
+# The point is for the user to remove these configuration records
+# one by one as the offenses are removed from the code base.
+# Note that changes in the inspected code, or installation of new
+# versions of RuboCop, may require this file to be generated again.
+
+# Offense count: 4
+Lint/Debugger:
+  Exclude:
```
MaximilianMeister: in velum we do the exludes inline, with the magic `rubocop:disable Lint/Debugger` comments around offensive areas, i am not sure which one is better, but we should stick to one way. do you have an opinion about what is better? global exlude or local?

velum-bootstrap/.rubocop_todo.yml
MalloZup: yop np

velum-bootstrap/.rubocop_todo.yml
MalloZup: from my POV i tend to like global ones for 2 reason:

- less lines everywhere (even if comments) but we are speking from 2 comment pro violation.
- with global you have better overview and can be maintained better with the command ```rubocop auto-gen-config```

For fixing the cops normally i just remove some offenses and try to fix them.

The only disadvantage with global is that you loose the offensive areas but this are  visible again when people start to fix the existing offenses(removing from todo.yml).

I would propose for global which spam less, but maybe i am missing some advantage for the inlines one :penguin: :thought_balloon:

MaximilianMeister: gotcha. i also like it, but i can see one drawback. the inline ones are more fine grained. luckily in this project the codebase is not so huge, so i would even say it doesnt matter because we only have specs and no ruby code to test against.

in velum i think inline and fine grained is useful to make some exceptions in e.g. complex methods to retain some readability instead of trying to shoehorn all future method definitions into a maximum of X lines. this would make most developers unhappy (or even angry and they will start to hate rubocop :P)

i'd also be fine to use a global config in this project here, and keep inline in velum because of the rather large codebase we're running rubocop against

MalloZup: @ereslibre @vitoravelino @mssola 

Question is : global exlude or local?

ereslibre: I don't have a strong opinion either way. What I like about the inline ones is that it bothers you, and you have two options: either you fix it, or you tell rubocop explicitly that you know what you are doing (you had the chance to choose). Having a global exclusion doesn't give you that option while you are working on something. My 2 cents.

MaximilianMeister: :+1: let's go with it and if it bothers us down the road we can easily drop the global excludes later

PR #368, opened 2018-05-04
Patch init containers too, not only regular containers

ereslibre:
Fixes: bsc#1091843


ereslibre:
[Bug 1091843 - When updating from 2.0 the `update all nodes` link disappears](https://trello.com/c/DnZfae7N/842-bug-1091843-when-updating-from-20-the-update-all-nodes-link-disappears)

PR #370, opened 2018-05-07
when there's only one master/worker subtract it from removables

MaximilianMeister:

@vitoravelino @inercia can you check if this logic is sufficient?

```
if there is only 1 master
  it's not removable
else
  all nodes are potentially removable
```

or is there more logic behind?

### DIFF
--- a/velum-bootstrap/spec/features/07-node-remove.rb
+++ b/velum-bootstrap/spec/features/07-node-remove.rb
@@ -4,6 +4,22 @@ require "yaml"
 feature "Remove a Node" do
 
   let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
+  let(:master_node_removable?) do
+    environment["minions"].count { |element| element["role"] == "master" }.tap do |count|
+      !count.zero? && count != 1
+    end
+  end
+  let(:worker_node_removable?) do
+    environment["minions"].count { |element| element["role"] == "worker" }.tap do |count|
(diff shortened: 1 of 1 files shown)


vitoravelino:
If there's only 1 master, this master is not removable.
If there's only 1 worker, this worker is not removable.

We only allow the removal of each node type if there's at least 2 of them.

MaximilianMeister:
@inercia @vitoravelino updated, thanks

velum-bootstrap/spec/features/07-node-remove.rb
```diff
@@ -4,6 +4,14 @@
 feature "Remove a Node" do
 
   let(:node_number) { environment["minions"].count { |element| element["role"] != "admin" } }
+  let(:master_node_removable?) { environment["minions"].count { |element| element["role"] == "master" } == 1 }
+  let(:worker_node_removable?) { environment["minions"].count { |element| element["role"] == "worker" } == 1 }
```
vitoravelino: Did you mean `not_removable`?

MaximilianMeister: yes indeed ...

MaximilianMeister: i kept the name and changed the logic for readability, a `not` check would have been more simple, but less readable `unless not_removable` sounds awkward

PR #372, opened 2018-05-07
KVM: Simplify libvirt pool creation steps

kiall:


caasp-kvm/README.md
```diff
@@ -70,18 +70,8 @@ git clone git@github.com:kubic-project/automation.git
...
-  <name>default</name>
-  <target>
-    <path>/path/to/libvirt/images</path>
-  </target>
-</pool>
-```
-
-```
-sudo virsh pool-create default-pool.xml
+sudo virsh pool-define-as --target /var/lib/libvirt/images/ --name default --type dir
```
stefsuse: `sudo virsh pool-define default-pool.xml` is the equivalent using a file, could be good to leave in as an alternative

ereslibre:
@stefsuse Are you OK with merging? I see you had a comment but not sure if you expect further changes :)

PR #373, opened 2018-05-08
check basename of image

MaximilianMeister:
it can be that a full path contains kubic


flavio:
Tested locally, it works.

PR #378, opened 2018-05-09
[3.0] Apply local changes to the haproxy config using cloud-init

ereslibre:
Otherwise, local modifications done to the configuration of haproxy
never applies when we create a development environment. It's kind of
a corner case, but can be confusing if you are messing with the
`haproxy` configuration and creating development environments.

(cherry picked from commit cc6dfbed52102faf73e53a877e9768404d44d28e)


ereslibre:
[Bug 1071994 - kubectl run --attach exits early](https://trello.com/c/nNNdxJAn/704-bug-1071994-kubectl-run-attach-exits-early)

PR #380, opened 2018-05-09
Add scaling parameters

FedericoCeratto:


ereslibre:
Wouldn't it allow us to override the number of masters/workers on the nightly job easily?, otherwise the defaults in https://github.com/kubic-project/jenkins-library/blob/master/vars/coreKubicProjectPeriodic.groovy#L19-L20 would always apply, no?

PR #381, opened 2018-05-09
fix node removal logic

MaximilianMeister:


MaximilianMeister:
> We got tricked with all the naming, uh? xD

:laughing:

PR #398, opened 2018-05-24
Move away from Virtualization:containers

kiall:
Virtualization:containers has way more than we really need in it, including
massively out of date packages which are now causing problems. Move to a new
Devel:CASP:CI project which contains just what we need.


kiall:
All of devenv's dependencies now install on Leap 15, and - likely - on Tumbleweed correctly.

PR #399, opened 2018-05-28
some fixes for node-removal

MaximilianMeister:


testinfra/tests/test_kubernetes_master.py
```diff
@@ -59,7 +59,11 @@ def test_kubernetes_cluster(self, host):
             host.file("/tmp/cluster_info/nodes.json").content_string
         )
 
-        env_file = os.environ['ENVIRONMENT_JSON']
+        try:
```
bergmannf: An alternative way to do this is:

```python
env_file = os.environ.get("ENVIRONMENT_JSON", "../caasp-kvm/environment.json")
```

MaximilianMeister: much better, thanks. updated

PR #400, opened 2018-05-28
Workaround bsc#1091097: use the kubic hardware-x86_64 image.

bergmannf:
The hardware-x86_64 image provides the required 9p*.ko modules for mounting
folders from the host machine into the vms.


bergmannf:
While this PR will *fix* (kind of) the problem that the folder can be mounted it still won't make the `kubic` image work, as it is also missing `docker`.

PR #405, opened 2018-05-30
moar node removal fixes and some DRY cleanup

MaximilianMeister:
afterall there were still some flaws ... :crying_cat_face:

### DIFF
--- a/velum-bootstrap/spec/features/07-node-remove.rb
+++ b/velum-bootstrap/spec/features/07-node-remove.rb
@@ -2,20 +2,10 @@ require "spec_helper"
 require "yaml"
 
 feature "Remove a Node" do
-  let(:master_node_removable?) do
-    environment["minions"].count { |element| element["role"] == "master" }.tap do |count|
-      !count.zero? && count != 1
-    end
-  end
-  let(:worker_node_removable?) do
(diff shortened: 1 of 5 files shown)


velum-bootstrap/spec/features/07-node-remove.rb
```diff
@@ -2,21 +2,10 @@
...
-    environment["minions"].count { |element| element["role"] == "worker" }.tap do |count|
-      !count.zero? && count != 1
-    end
-  end
   let(:node_number_removable) do
     removable = node_number
-    removable -= 1 unless master_node_removable?
-    removable -= 1 unless worker_node_removable?
+    removable -= master_node_number unless node_removable?(role: "master")
+    removable -= worker_node_number unless node_removable?(role: "worker")
```
vitoravelino: Isn't these going to be > 1? The idea is to remove only one node, no?

vitoravelino: I mean `master_node_number` and/or  `worker_node_number`.

MaximilianMeister: the logic is: when any master or worker is not removable, we should remove all of them from the number of removable nodes

PR #414, opened 2018-06-05
Added test for firewalld status.

danielorf:
Ran into a problem when building with firewall disabled, so I added this check.


kiall:
Are you sure this is necessary? We don't have this enabled for CI runs (and we can see CI failed here due to this new check)

kiall:
Actually - I wonder if what you're really hitting is Dockers broken MTU handling?

Try adding the MTU flag in /etc/sysconfig/docker, restarting the docker daemon, and recreating any containers?

    DOCKER_OPTS="--mtu=1400"

MalloZup:
@danielorf  i would not add this on CI but rather on a doc/wikipage somewhere

MalloZup:
i don't think is needed, closing ! thank you for all the efford and feel free to come up with documentatoin or other solution. 
( i close this pr also because there was no reactions on comments from 5 months ago, i think in realistic/pragmatic way this PR become outdated, this is why i close it)

THX :rose:

PR #418, opened 2018-06-07
OpenStack Terraform: Support SES team's needs

BlaineEXE:
The SES team also needs to support deploying development and test CaaSP clusters on OVH. Heat is not an option, so we must use Terraform. Make the necessary changes to bring the Terraform orchestration up to feature parity with the Heat orchestration. Also make the changes to allow creation of clusters on OVH, which does not have network creation or floating IP capabilities.

I have done my best to make sure that the Terraform configuration out-of-the box will behave exactly the same as the QA team has maintained it thus far.


BlaineEXE:
@kiall @mjura Could you take a look at this an let me know if it looks good to merge?

PR #419, opened 2018-06-08
Silence virt-cat, update readme

FedericoCeratto:
Send virt-cat -h output do /dev/null
Improve README.md with development tips

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,11 @@ large scale clusters for testing and validation.
 A tool to build a OpenStack based CaaSP cluster. Well suited to building
 large scale clusters for testing and validation.
 
+## CaaSP Bare Metal
+
+A tool to build a Bare Metal cluster. Acts as a client for the Bare Metal
+Manager service.
+
 ## Jenkins Pipelines
 
 A set of Jenkins pipelines which are not directly associated with a specific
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -40,3 +45,21 @@ A set of tests to validate a CaaSP cluster has been successfully deployed
 
 Some scripting to walk through the Velum UI to bootstrap a cluster, primarily
 used within CI.
+
+# Development
+
+Where possible, use Python, Ruby, Bash.
```
stefsuse: I really like the idea of adding development guidelines, but the instructions in this section should be expanded to be more clear in what we would like to see.
 
An example:
"Prefer configuration paths" <--> "Avoid hard coding configuration data, such as file system locations, URIs, and certain defaults. Use separate configuration files."

misc-tools/extract-kernel-initrd-cmdline-args
```diff
@@ -52,10 +52,11 @@ def parse_args
 end
 
 def ensure_guestfs_tools_installed
-  if !system("virt-cat --help")
+  puts "Checking for virt-cat"
+  if !system("virt-cat --help > /dev/null")
```
stefsuse: If you have to change it twice consider making it a function

README.md
FedericoCeratto: There are few other suggestions in the change, including "CI logs can be difficult to inspect: avoid output that look like errors if unnecessary.". Printing out the output of the help command catches the eye of people reviewing the logs as if a command could not parse options and printed out its help.

PR #431, opened 2018-06-14
JJB: Change notification-context's to be something nicer to read

kiall:
These will end up rendered in GitHub as:

    Jenkins/Flake8/pr-head

etc instead of:

    continuous-integration/jenkins/Flake8/pr-head


dannysauer:
Still seems to work fine with current CI checks.  Any reason not to merge in the more human-readable names?  @MaximilianMeister @tdaines42 either of you have an opinion?

MalloZup:
@dannysauer this is basically for the jenkins-job builder which is triggered manually.

Changing the names there will not have consequences directly for CI. The job name will just change

PR #433, opened 2018-06-14
download-image: Find SHA256 by regex

BlaineEXE:
Looking for SHA256 sums in files by specific columns fails if the files'
columns are not laid out to the get_sha256 function's assumptions.
Instead, look for a length-64 hex word in the file text by regex to find
the sha256 sum. Error if no sum is found or if more than one word
matches the regex.

### DIFF
--- a/misc-tools/download-image
+++ b/misc-tools/download-image
@@ -119,7 +119,7 @@ def get_canonical_url(args, url):
     }
 
     image = requests.head(url, proxies=proxies)
-    
+
     if image.status_code == 302 or image.status_code == 301:
         return get_canonical_url(args, image.headers.get('Location'))
     elif image.status_code == 200:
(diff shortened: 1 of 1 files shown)


BlaineEXE:
@kiall @MaximilianMeister @bergmannf Are any of you available to review?

misc-tools/download-image
```diff
@@ -119,35 +119,44 @@ def get_canonical_url(args, url):
...
     if image.status_code == 302 or image.status_code == 301:
         return get_canonical_url(args, image.headers.get('Location'))
     elif image.status_code == 200:
         return url
     else:
         raise Exception("Cannot find image's real location: " + url)
 
+def find_sha256(text):
+    # Find the regex by looking for a length-64 hex word
+    sha_regex = re.compile(r'\b([0-9A-Fa-f]{64})\b')
```
bergmannf: Little nitpick: I don't think `re.compile` is needed, as the regular expression is not [reused](https://docs.python.org/3/library/re.html#re.compile).

Instead `re.findall(pattern, string)` could be used.

PR #439, opened 2018-06-22
[2.1] use older bundler

MaximilianMeister:
otherwise kubeclient wont be installed in the development image


MaximilianMeister:
i'll have to check if we need a fix on the 2.0 branch as well

MaximilianMeister:
`release-2.1` is out of maintenance (EOL) already

PR #440, opened 2018-06-22
Upgrade job: Update to use both stages

kiall:
This allows for the job to change branch where necessary. Another later
cleanup will remove the hardcoded branch name. I'm thinking adding
the hardcode is OK given we already hardcode the update repo URL a few
lines above.

### DIFF
--- a/jenkins-pipelines/Jenkinsfile.kubic-nightly-upgrade
+++ b/jenkins-pipelines/Jenkinsfile.kubic-nightly-upgrade
@@ -34,8 +34,19 @@ coreKubicProjectPeriodic(
     )
 
     // Run through the upgrade orchestration
-    upgradeEnvironment(
-        environment: environment
+    upgradeEnvironmentStage1(
+       environment: environment
+    )
+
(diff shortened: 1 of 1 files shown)


jenkins-pipelines/Jenkinsfile.kubic-nightly-upgrade
```diff
@@ -34,8 +34,19 @@ coreKubicProjectPeriodic(
     )
 
     // Run through the upgrade orchestration
-    upgradeEnvironment(
-        environment: environment
+    upgradeEnvironmentStage1(
+       environment: environment
+    )
+
+    // On master, we have no branch to switch to..
```
FedericoCeratto: Perhaps add "When running on other branches we switch to a newer branch to test an upgrade"

kiall: Yep - agree. I intend to move these things into a `misc-files/upgrade-urls.json` or something similar ASAP though too, so all the places that need changing are together in 1 place.

PR #451, opened 2018-07-17
node-remove: don't disable meltdown/spectre fixes

MaximilianMeister:
this is also done in the node-add job

**cc @kiall** without this, the bootstrap will always fail


MaximilianMeister:
node removal job is green in the meantime

PR #457, opened 2018-07-24
addNode: reduce worker number

MaximilianMeister:
as we have increased the admin ram in CI we need to scale down
on the addNode job

Error: cannot set up guest memory 'pc.ram': Cannot allocate memory


MaximilianMeister:
this is more of a wild guess, i hope this is fixing it

an example error can be found here: http://jenkins.caasp.suse.net/job/caasp-nightly-node-add/job/master/134/console

MaximilianMeister:
triggered a test run via replay http://jenkins.caasp.suse.net/job/caasp-nightly-node-add/job/master/135/

PR #462, opened 2018-08-03
adapt to the new sonobuoy workflow

MaximilianMeister:


MaximilianMeister:
let's merge it now to fix the tests (they time out) we can send follow ups later. tested locally

PR #464, opened 2018-08-08
Add deploy option to add CAP secgroup

HartS:
This introduces a `--include-cap-ports` option to the `caasp-openstack` deploy script.

When the deploy script is run with this option, the worker nodes will be associated with a security group which allows inbound access on ports required for CAP functionality.

When run without the option, the worker nodes will be associated with a 'no-op' security group which doesn't provide any additional rules.

### DIFF
--- a/caasp-openstack-heat/caasp-openstack
+++ b/caasp-openstack-heat/caasp-openstack
@@ -7,6 +7,7 @@ HAS_ACTION=false
 RUN_BUILD=false
 RUN_UPDATE=false
 RUN_DESTROY=false
+INCLUDE_CAP_PORTS=off
 
 NAME=caasp-stack
 OPENRC_FILE=
@@ -25,6 +26,8 @@ Usage:
     -m|--masters             <INT>   Number of masters to build (Default: 3)
     -w|--workers             <INT>   Number of workers to build (Default: 2)
     -i|--image               <STR>   Image to use
+    --include-cap-ports              Open additional ports for testing/deployment of SUSE Cloud Application
+                                     Platform (SUSE Cloud Foundry)
 
   * Destroying a cluster
 
@@ -103,6 +106,8 @@ while [[ $# -gt 0 ]] ; do
       echo "$USAGE"
       exit 0
       ;;
+    --include-cap-ports)
+      INCLUDE_CAP_PORTS=on
   esac
   shift
 done
@@ -124,7 +129,8 @@ build_stack() {
   openstack stack create --verbose --wait -e "$HEAT_ENVIRONMENT_FILE" -t caasp-stack.yaml "$NAME" \
     --parameter master_count=$MASTERS \
     --parameter worker_count=$WORKERS \
-    --parameter image="$IMAGE"
+    --parameter image="$IMAGE" \
+    --parameter include_cap_ports=$INCLUDE_CAP_PORTS
 
   log "CaaSP Stack Created with name $NAME"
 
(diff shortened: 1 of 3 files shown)


caasp-openstack-heat/caasp-openstack
```diff
@@ -25,6 +26,7 @@ Usage:
     -m|--masters             <INT>   Number of masters to build (Default: 3)
     -w|--workers             <INT>   Number of workers to build (Default: 2)
     -i|--image               <STR>   Image to use
+    --include-cap-ports              Open additional ports for testing/deployment of CAP
```
mjura: Can you please expand this name CAP - Cloud Application Platform (Cloud Foundry)

caasp-openstack-heat/caasp-stack.yaml
```diff
@@ -84,6 +85,10 @@ parameters:
     type: string
     description: Root Password for the VMs
     default: linux
+  include_cap_ports:
+    type: string
+    description: Add cap-specific security group to worker nodes
+    default: "false"
```
mjura: Instead `false` and `true` which are boolean values, please use `enable|disable` or `on|off`

caasp-openstack-heat/caasp-openstack
HartS: Sure, how does changing the comment to `Open additional ports for testing/deployment of SUSE Cloud Application Platform (Cloud Foundry)` sound

HartS:
Added port 2793 to this list

MaximilianMeister:
CI was halted, job is running now. lets wait for the result

jimmykarily:
This PR assumes that the external-ips value in scf-config-values will be one of the workers. If the user sets that to the IP of a master node then the services will not be reachable because the ports are only opened on the worker nodes.

A possible solution would be to open the ports on all nodes. The other would be to specify on which node you want the cap ports to be open and the user should then make sure she uses that node's IP in the scf-config-values.yaml file.

jimmykarily:
I'm currently trying out a different approach here: https://github.com/SUSE/caasp-openstack-heat-templates/compare/master...jimmykarily:add_option_for_cap_secgroup

I'm checking if all options work as expected.

jimmykarily:
Better link: https://github.com/SUSE/caasp-openstack-heat-templates/pull/10/files

caasp-openstack-heat/caasp-stack.yaml
```diff
@@ -140,6 +145,34 @@ resources:
...
+  secgroup_cap_worker_on:
+    type: OS::Neutron::SecurityGroup
+    properties:
+      rules:
+        - protocol: tcp
+          port_range_min: 2222
+          port_range_max: 2222
+        - protocol: tcp
+          port_range_min: 2341
+          port_range_max: 2341
```
jimmykarily: This port is not needed. It was a relic from some debugging session which made it to the docs. Please remove.

HartS:
@jimmykarily :
> This PR assumes that the external-ips value in scf-config-values will be one of the workers.

That's what we already expect. QA has been manually adding a floating IP to one of the workers, but we should probably have the HEAD template do this when `--include-cap-ports` is used

HartS:
@jimmykarily Port 2341 has been removed now

MalloZup:
any update on this PR why is on blocked? can we merge it? do we need it? tia :sunflower:

MalloZup:
stale, feel free to repopen

PR #467, opened 2018-08-14
Check if sonobuoy binary is present and executable

mjura:


k8s-e2e-tests/e2e-tests
```diff
@@ -50,7 +50,7 @@ while [[ $# > 0 ]] ; do
 done
 
 run_tests() {
-    if ! which sonobuoy; then
+    if command -v sonobuy > /dev/null; then
```
MaximilianMeister: sonobuoy

mjura: Done

k8s-e2e-tests/e2e-tests
```diff
@@ -50,7 +50,7 @@ while [[ $# > 0 ]] ; do
 done
 
 run_tests() {
-    if ! which sonobuoy; then
+    if command -v sonobuoy > /dev/null; then
```
MaximilianMeister: it needs to be `if ! command -v sonobuoy > /dev/null`

PR #477, opened 2018-08-29
Add parallel ssh

FedericoCeratto:
Add tool to run ssh in parallel and stream output in realtime.
Add helper to install Python virtualenv and dependencies.


ereslibre:
@FedericoCeratto is in the scope of this PR to also integrate this tool with our current pipelines?

PR #478, opened 2018-08-31
Add pipelines and updated e2e-tests to run upstream kubernetes e2e tests

tdaines42:


tdaines42:
Related to https://github.com/kubic-project/jenkins-library/pull/212

tdaines42:
[(3) Enable Kubernetes e2e tests](https://trello.com/c/EaTLBKDB/1136-3-enable-kubernetes-e2e-tests)

jenkins-pipelines/Jenkinsfile.kube-e2e-nightly
```diff
@@ -0,0 +1,29 @@
...
+    environmentTypeOptions: kvmTypeOptions,
+    environmentDestroy: env.ENVIRONMENT_DESTROY.toBoolean(),
+) {
+    // empty preBootstrapBody
+} {
+    stage('Run K8S E2E Tests') {
+        runK8sE2eTests(
+            environment: environment,
+            e2eFocus: '',
+            e2eSkip: 'Alpha|\\[(Conformance|Disruptive|Feature:[^\\]]+|Flaky|Serial|Slow)\\]'
```
mjura: Why do you want to skip Serial tests? Default here is
```
"Alpha|Kubectl|\\[(Disruptive|Feature:[^\\]]+|Flaky)\\]"
```

jenkins-pipelines/Jenkinsfile.kube-e2e-serial-nightly
```diff
@@ -0,0 +1,29 @@
...
+coreKubicProjectPeriodic(
+    environmentTypeOptions: kvmTypeOptions,
+    environmentDestroy: env.ENVIRONMENT_DESTROY.toBoolean(),
+) {
+    // empty preBootstrapBody
+} {
+    stage('Run K8S E2E Serial/Disruptive Tests') {
+        runK8sE2eTests(
+            environment: environment,
+            e2eFocus: '\\[(Serial|Disruptive)\\]',
```
mjura: Is there any upstream explanation for this?

k8s-e2e-tests/e2e-tests
```diff
@@ -61,8 +72,14 @@ run_tests() {
     if ! command -v sonobuoy > /dev/null; then
         abort "sonobuoy needs to be installed"
     fi
+
     test_config
-    sonobuoy run --kubeconfig $KUBECONFIG --config sonobuoy.json
+
+    if [ -n "${E2E_FOCUS}" ] || [ -n "${E2E_SKIP}" ]; then
```
mjura: We can skip this `if` and rely on default values `E2E_FOCUS=''` and `E2E_SKIP=''`

tdaines42:
The conformance tests are still as far as I know all that's needed for certification. There are more tests upstream that we can use to test the functionality of our product. So, I was asked by @kiall  to find a way to get those tests running.

As far as the filtering for the different pipelines based on upstream info the serial tests can’t be ran in parallel with other tests. They are being filtered so they can be ran in a separate pipeline. 

There’s more info on the tests and how and why they run this way here:
https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md#kinds-of-tests

and here: 
https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md#continuous-integration

k8s-e2e-tests/e2e-tests
tdaines42: Using `E2E_FOCUS=''` and `E2E_SKIP=''` would actually enable all the tests which is something we don't want. I will find a better solution for this.

mjura:
Ok, I have found everything in 
https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md#what-runs-in-ci

My another thought on this is that some tests will be failing always on our CI due to poor performance. I don't know if it make sense to define these tests if they will be broken permanently

tdaines42:
I believe Kiall's idea was to get these running then disable any tests that we don't feel are applicable or are broken.

PR #486, opened 2018-09-11
Fixes test case that is looking for outdated text in UI

chentex:
Integration test are not passing in https://github.com/kubic-project/velum/pull/647 because of a changed text in the UI form this PR is required for the other one to run. And probably will make this build fail until the other one is in place.


chentex:
[Bug 1099985 -  remove unassigned nodes from the cluster](https://trello.com/c/YVxD1zmo/1149-bug-1099985-remove-unassigned-nodes-from-the-cluster)

PR #487, opened 2018-09-11
bootstrap: rely on global orchestration indicator

vitoravelino:
So far we were relying on the number of green/red nodes to determine if
the orchestration had succeeded or not. Since we added a global
orchestration indicator to the screen, we can simplify this check by
just checking it.

Signed-off-by: Vítor Avelino <contact@vitoravelino.me>

Depends on https://github.com/kubic-project/velum/pull/653


MalloZup:
feel free tp reopen

PR #492, opened 2018-09-17
Move virtualenv to workspace

FedericoCeratto:


caasp-bare-metal/deployer/deployer
```diff
@@ -1,9 +1,4 @@
...
-type python3 >/dev/null 2>&1 || sudo zypper install --no-confirm python3
-
-if [ ! -d ~/py3venv ]; then
-  python3 -m venv ~/py3venv
-  ~/py3venv/bin/pip3 install PyYAML prometheus_client urllib3 robobrowser
-fi
-~/py3venv/bin/python3 ./deploy_testbed.py $@
+set -u
+../../misc-tools/setup-python PyYAML prometheus_client urllib3 robobrowser
+${WORKSPACE:-~}/py3venv/bin/python3 ./deploy_testbed.py $@
```
kiall: Why use `${WORKSPACE:-~}` vs just putting it in the pwd as our other tools do? Having it in the same place for both CI runs, and local runs, makes it easier to understand + grok. It's a small "personal taste" thing, so I wouldn't hold up the merge on it though.

FedericoCeratto: Using the current directory requires the Jenkins script to have a cd(...) block around every command and any inconsistency across groovy scripts can lead to tricky errors.

PR #498, opened 2018-10-01
Do not install packages during CI

drpaneas:
The installation of `$DIST_PACKAGES` and `$EARLY_DIST_PACKAGES` happens
only outside of the CI, so the Jenkins base image should come
pre-baked with the correct packages. Any installation attempt that
happens during the CI should fail.

### DIFF
--- a/caasp-devenv
+++ b/caasp-devenv
@@ -239,8 +239,7 @@ export ENVIRONMENT="$DIR/caasp-kvm/environment.json"
 # Core methods
 setup() {
   log "Installing CaaSP Development Environment Requiemnts"
-  # || : is necessary, as Zypper exits non-zero for "no changes".
-  sudo zypper in --no-confirm $EARLY_DIST_PACKAGES || :
+  sudo zypper in --no-confirm --auto-agree-with-licenses $EARLY_DIST_PACKAGES
 
   local dist=$(lsb-release -sd | tr -d '"' | tr " " "_")
 
@@ -262,8 +261,7 @@ setup() {
     sudo zypper --gpg-auto-import-keys ref systemsmanagement_terraform
   fi
 
-  # || : is necessary, as Zypper exits non-zero for "no changes".
-  sudo zypper in --no-confirm $DIST_PACKAGES || :
+  sudo zypper in --no-confirm --auto-agree-with-licenses $DIST_PACKAGES
 
   # Enable and start services, if disabled
   sudo systemctl enable --now docker
--- a/velum-bootstrap/velum-interactions
+++ b/velum-bootstrap/velum-interactions
@@ -68,6 +68,25 @@ warn()       { log "WARNING: $@" ; }
 error()      { log "ERROR: $@" ; exit 1 ; }
 check_file() { if [ ! -f $1 ]; then error "File $1 doesn't exist!"; fi }
 
+# Test methods
+is_ci()
+{
+    if [ -e /boot/grub2/grub.cfg ] && grep 'jenkins-worker' /boot/grub2/grub.cfg > /dev/null; then
+        echo "This image is generated for Jenkins CI purposes."
+        return
+    fi
+    false
+}
+
+# Velum interaction packages
+which ruby || error "Ruby is not installed"
+ruby_version=$(ruby --version | cut -d ' ' -f2 | cut -d '.' -f1-2)
+INTERACTION_PACKAGES="ruby${ruby_version}-rubygem-bundler \
+                      ruby${ruby_version}-devel \
+                      phantomjs \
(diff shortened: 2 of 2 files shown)


caasp-devenv
```diff
@@ -136,6 +136,19 @@ error()      { log "ERROR: $@" ; exit 1 ; }
 check_file() { [ -f "$1" ] || error "File $1 doesn't exist!" ; }
 usage()      { echo "$USAGE" ; exit 0 ; }
 
+# Test methods
+check_conflicts()
+{
+    log "Checking Dependencies"
+    eval local  packages='$'$1
```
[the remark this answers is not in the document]
drpaneas: done ;)

caasp-devenv
```diff
@@ -239,8 +252,7 @@ export ENVIRONMENT="$DIR/caasp-kvm/environment.json"
 # Core methods
 setup() {
   log "Installing CaaSP Development Environment Requiemnts"
-  # || : is necessary, as Zypper exits non-zero for "no changes".
-  sudo zypper in --no-confirm $EARLY_DIST_PACKAGES || :
+  check_conflicts EARLY_DIST_PACKAGES && sudo zypper in --no-confirm $EARLY_DIST_PACKAGES
```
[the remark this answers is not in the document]
drpaneas: I agree

caasp-devenv
```diff
@@ -136,6 +136,19 @@ error()      { log "ERROR: $@" ; exit 1 ; }
...
 
+# Test methods
+check_conflicts()
+{
+    log "Checking Dependencies"
+    local  packages="$*"
+    if sudo zypper in --no-confirm --auto-agree-with-licenses --dry-run $packages > /dev/null 2>&1 ; then
+        echo -e "Dependencies are fine.\n"
+    else
+        sudo zypper in --no-confirm --auto-agree-with-licenses --dry-run $packages
```
bergmannf: This is just running the `zypper` command a second time isn't it?

Given that the first run's return code is used in the `if` statement.

drpaneas: Yes. The first time it's not visible (sent to /dev/null) and it determines if there's a problem or not. In case there is a problem then I am running it again but this time it is visible (for debugging purposes). My point is that I didn't want to pollute the screen in case everything goes well.

But I can always redirect the output into a file. Let me do that :)

dannysauer: Ideally, don't redirect to to a temp file when you can keep it in memory and check the return code with `$?`. :)
```shell
output=$(sudo zypper in --no-confirm --auto-agree-with-licenses --dry-run $packages 2>&1)
if [ $? -eq 0 ]
```

drpaneas:
Since a few hours ago the CI environment has changed. Although yesterday there were no repositories, today there's a repo installed:

```bash
prvcld-leap-15:~ # zypper lr -U
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias           | Name                    | Enabled | GPG Check | Refresh | URI                                                                   
--+-----------------+-------------------------+---------+-----------+---------+-----------------------------------------------------------------------
1 | network_cluster | Clustering (SLE_12_SP3) | Yes     | (r ) Yes  | No      | http://download.opensuse.org/repositories/network:/cluster/SLE_12_SP3/
```

Is this intentional? If it's not, and there should not be any repos in the CI environment, should I include a check for that in the script? The current version of this PR is assumes that a system with no repositories is the CI environment.

caasp-devenv
drpaneas: Great suggestion @dannysauer  let me do that ;)

kiall:
@drpaneas Yea, that repo really shouldn't be there. Adding repos tends to lead to getting a really odd mix of packages - and things that aren't the official Leap repos tend to have breakages ALL THE TIME.

On top of that, we shouldn't be installing any extra packages at runtime. Everything should be included in the base image, so that on days like today when software.opensuse.org is down, our CI isn't entirely blocked.

https://github.com/kubic-project/automation/pull/501 removes the code that installs this repo.

https://build.suse.de/package/rdiff/Devel:CASP:CI/jenkins-worker-image-leap15.0?linkrev=base&rev=56 adds the pssh package to the base image.

cc @FedericoCeratto

caasp-devenv
dannysauer: Whoops; forgot about `set -e`.

If you really wanted the output, you could add a `set +e` in the function right before the declaration. I _think_ calling a function `f` like `f && :` also implicitly treats everything in the function like it's tested as well, so has the same effect of locally disabling error-exit - but that might be Bash-specific.

drpaneas:
@kiall thanks for the explanation. Is there anything else needed for this to be merged?

PR #501, opened 2018-10-03
Remove usage of SLE Repo on Leap

kiall:
We should not be installing extra repos during CI, certainly not SLE12 repos on
Leap! Additionally,  we should avoid installing packages at all during CI and
instead bake them into the base Jenkins image.


kiall:
See also https://build.suse.de/package/rdiff/Devel:CASP:CI/jenkins-worker-image-leap15.0?linkrev=base&rev=56

PR #508, opened 2018-10-17
Move container-feeder to an optional test.

bergmannf:
Container-feeder is only needed, when the node is not running against a registry. This should only be the case for SLE12.

If the node does not use registry-image (use as default case here) the container-feeder is expected to be running.

To test correctly against `SLE15` images, the images need to include a `caasp-container-manifests` package with https://github.com/kubic-project/caasp-container-manifests/pull/203 already merged.


MalloZup:
@bergmannf can we merge this after conflicts done?

testinfra/tests/utils.py
```diff
@@ -28,5 +29,21 @@ def environment():
...
+
         return set(feature_data.items()).issubset(set(value.items()))
+
+    @staticmethod
+    def load_registry_configuration(host):
+        try:
+            registry_config = host.file(
+                '/usr/share/caasp-container-manifests/config/registry/registry-config.yaml'
+            ).content_string
+        # unfortunately if the file doesn't exist it will throw a RuntimeError
```
MalloZup: +1 for using comments in good way

PR #512, opened 2018-10-19
Allow SLE15 as devenv

bergmannf:
This PR adds a new channel `devel_15`, that will use the new `qcow2` images for SLE15 and patch the container manifests to use docker-images from `registry.suse.de`.

### DIFF
--- a/caasp-kvm/tools/fix-kubelet-manifest
+++ b/caasp-kvm/tools/fix-kubelet-manifest
@@ -20,6 +20,14 @@ opts_parser = OptionParser.new do |opts|
     opts.on("-k", "--kubic", "Patch image names from sles to kubic") do |k|
       options[:kubic] = k
     end
+
+    opts.on("--sle15", "Patch image sources to registry") do |f|
+      options[:sle15] = f
+    end
+
+    opts.on("--noinit", "Do not perform initContainer patching") do |n| 
+      options[:noinit] = n
+    end
 end
 opts_parser.parse!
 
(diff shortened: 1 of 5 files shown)


caasp-kvm/tools/fix-kubelet-manifest
```diff
@@ -101,6 +121,13 @@ if options[:kubic]
   manifest = YAML.load(YAML.dump(manifest).gsub!("sles12", "kubic"))
 end
 
+if options[:sle15]
+  manifest = YAML.safe_load(YAML.dump(manifest).gsub(/sles12\/(.*):__TAG__/, 'registry.suse.de/devel/casp/3.0/controllernode/images_container_base/sles12/\1:__TAG__'))
```
MalloZup: nitpickL we might create a var for `registry.suse.de/devel/casp/3.0/controllernode/images_container_base/sles12/\1:__TAG__`?

i am thinkink when we will just add th `4.0` :smile:

bergmannf: @MalloZup We could add a `variable` to make it more readable.

It would probably be even better to read that value from the `https://github.com/kubic-project/caasp-container-manifests/blob/master/config/registry/registry-config.yaml` if that contains valid values though.

bergmannf: @MalloZup I realized that if we use the values that the user has on his harddisk it could break expectations, when running `vanilla`, so maybe for now I would just leave the `PR` as-is until someone runs into problems.

Otherwise the `vanilla` configuration would read values from the repo to determine patch things that would later not be used at all.

misc-tools/download-image
```diff
@@ -252,7 +255,10 @@ def get_channel_url(args):
     if args.type == "docker":
         base_url += 'images_container_base'
     elif args.type in ["kvm", "openstack"]:
-        base_url += 'images'
+        if args.url == "channel://devel_15":
+            base_url += 'images-sle15'
+        else:
+            base_url += 'images'
```
MaximilianMeister: MINOR: could be also reduced to 2 lines:

```python
base_url += 'images'
if args.url == "channel://devel_15": base_url += '-sle15'
```

bergmannf: Update to use `3` lines though - more pythony IMO

PR #514, opened 2018-10-29
only ping on build

dannysauer:
I'm pretty sure the only time VPN connectivity is needed is during build, and it's mildly aggravating when destroy complains that it can't ping. :)

### DIFF
--- a/caasp-kvm/caasp-kvm
+++ b/caasp-kvm/caasp-kvm
@@ -255,7 +255,7 @@ if [[ $(basename $IMAGE) =~ [Kk]ubic ]]; then
   DIST=kubic
   KUBIC_MANIFEST_FLAG="--kubic"
 # check if internal network is reachable and fail early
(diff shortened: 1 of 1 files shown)


dannysauer:
LOL.  The CI build I triggered a moment ago with a commit in kubic-project/velum#665 just failed because it couldn't destroy the test environment when the ping failed. :D

MalloZup:
@dannysauer can we merge it?

dannysauer:
It works properly for me, so _I_ think it's safe to merge.

MalloZup:
I think look good but maybe there is a missing know-how on prod? Dunno i am no familiar with it already. Adding some review so we canare merge it with no pb

dannysauer:
I should better explain, then: The ping is there already, so the script can fail early if there's no network connectivity. As far as I can tell, host-external network connectivity is only needed if the build arg is provided; all other procedures happen locally (or technically via networking with the local KVM). So this limits the network check to only happen before doing things that actually need the network.

PR #516, opened 2018-10-30
Add caasp-kvm release image to volume delete list

dannysauer:
Leaving the "suse-caasp-whatever" KVM image in the storage volume seems to "sometimes" prevent Terraform from assembling the cluster.  So, add those images to the list to remove in manual cleanup script.
Also:
 * cleaned up the grep regex using a few `-e` options to make it a bit easier to read (and extend in the future)
 * removed the unnecessary shell wrapper in the xargs command


suseclee:
LGTM.

PR #112, opened 2017-09-04
Velum Bootstrap Cleanups

kiall:
Add some minor velum bootstrap cleanups. See commit messages for details.

PR #517, opened 2018-10-30
Remove global function and adapt smaller context

MalloZup:
The helper function was only used 1 time

this is a backport from a old PR

https://github.com/kubic-project/automation/pull/112

### DIFF
--- a/velum-bootstrap/spec/features/01-setup-velum.rb
+++ b/velum-bootstrap/spec/features/01-setup-velum.rb
@@ -14,7 +14,15 @@ feature "Register user and configure cluster" do
 
   scenario "User registers" do
     with_screenshot(name: :register) do
-      register
+      puts ">>> Registering user"
+      with_status_ok do
+        visit "/users/sign_up"
+      end
+      fill_in "user_email", with: "test@test.com"
+      fill_in "user_password", with: "password"
+      fill_in "user_password_confirmation", with: "password"
+      click_on "Create Admin"
+      puts "<<< User registered"
     end
   end
 
@@ -61,6 +69,15 @@ feature "Register user and configure cluster" do
 
       click_on "Next"
       puts "<<< Velum set up"
+    end
+  end
+
+  scenario "User proceeds past AutoYaST page" do
+    with_screenshot(name: :skip_autoyast) do
+      puts ">>> Skipping past the AutoYaST page"
+      visit "/setup/worker-bootstrap"
+      click_on "Next"
+      puts "<<< Skipped past the AutoYaST page"
     end
   end
 end
--- a/velum-bootstrap/spec/support/helpers.rb
+++ b/velum-bootstrap/spec/support/helpers.rb
@@ -26,19 +26,6 @@ module Helpers
     puts "<<< User logged in"
   end
 
-  def register
-    puts ">>> Registering user"
-    with_status_ok do
-      visit "/users/sign_up"
-    end
-
-    fill_in "user_email", with: "test@test.com"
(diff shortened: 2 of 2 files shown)


MalloZup:
@MaximilianMeister CC

velum-bootstrap/spec/features/01-setup-velum.rb
```diff
@@ -14,7 +14,16 @@
 
   scenario "User registers" do
     with_screenshot(name: :register) do
-      register
+     puts ">>> Registering user"
```
MaximilianMeister: one more space for indentation missing for the whole block of code below

velum-bootstrap/spec/features/01-setup-velum.rb
```diff
@@ -14,7 +14,16 @@
...
+     with_status_ok do
+       visit "/users/sign_up"
+     end
+     fill_in "user_email", with: "test@test.com"
+     fill_in "user_password", with: "password"
+     fill_in "user_password_confirmation", with: "password"
+     click_on "Create Admin"
+     puts "<<< User registered"
+
```
MaximilianMeister: empty line could be dropped

MaximilianMeister:
@MalloZup shall we add the other changes too from https://github.com/kubic-project/automation/pull/112 ?

MalloZup:
@MaximilianMeister thx :love_hotel: 3 things:

1) was thinking it run `rubocop` or `rubocop-rspec` via travis.( this was also why i did not running nothing locally).  Why is not the case?
___
2) for the additional changes : 
*   `configure` method was removed already by this commit https://github.com/kubic-project/automation/commit/df774e7b014186ad3da3acf079ceb118efc9207e#diff-611e4142df421de6a9c053afa84ae6d8

So what is left is this test ( i left over because i was not sure)
https://github.com/kubic-project/automation/pull/112/files#diff-847482c7cf57b6afd2982900bd75831dR40

I think we can add it , but it is skipping the autoyast test, it looked more as a hack to me. what is your POV there?
___
3) should i backport this to 30? should i backport any PR to 30 or other branches? What is the golden rule there :dog: 

Now some emoticons: :japan: :factory: :jack_o_lantern: :aerial_tramway:

MaximilianMeister:
> 1. was thinking it run `rubocop` or `rubocop-rspec` via travis.( this was also why i did not running nothing locally).  Why is not the case?

no real reason, we could have some style checks at some point, e.g. same rubocop checks as we have in velum. but we have also many different tools in automation, so it's quite some effort. i guess we didn't think it's important enough yet.

> 
> So what is left is this test ( i left over because i was not sure)
> https://github.com/kubic-project/automation/pull/112/files#diff-847482c7cf57b6afd2982900bd75831dR40
> 
> I think we can add it , but it is skipping the autoyast test, it looked more as a hack to me. what is your POV there?

IMO we should add it, because it's more of a hack **not** to go through the autoyast page. in the velum UI we have this page which has only static information text, so we skipped it until now

> 
> 1. should i backport this to 30? should i backport any PR to 30 or other branches? What is the golden rule there dog

@FedericoCeratto creates some batched commit backports from time to time

MalloZup:
for rubocop, i created already time ago this https://github.com/kubic-project/automation/blob/master/.travis.yml  
and we can see here that it run only for spec https://travis-ci.org/MalloZup/automation


For the autoyast i will add it thx ! :tokyo_tower:

MalloZup:
for the backport commits, i would like to know more about it , especially how is really the workflow there :smiley_cat:

PR #490, opened 2018-09-17
Adapt cloudinit api change to reflect new api in terraform-libvirt

MalloZup:
### What do this PR?

in the v0.5.00 of `terraform-libvirt-plugin` the API for the `libvirt_cloudinit` will change to `libvirt_cloudinit_disk`.

This PR will adapt the kubic-terraform tf files to reflect this change.

### What is needed for this PR to be merged?

- [x]  Upstream PR get merged https://github.com/dmacvicar/terraform-provider-libvirt/pull/410
- [x] test locally this PR if it works
- [x]  In kubic-ci you need the latest plugin to be installed

### additional info:

with the 0.5.0 version doing multiples time `terraform plan` with a `cloud_init_disk` resource doesn't produce anymore any resource creation for nothing, see issue https://github.com/dmacvicar/terraform-provider-libvirt/issues/313

PR #519, opened 2018-10-30
[Release-30]: adapt cloudinit api change to

MalloZup:
backport from : https://github.com/kubic-project/automation/pull/490

cloud_init_disk to reflect upstream api

### DIFF
--- a/caasp-kvm/cluster.tf.erb
+++ b/caasp-kvm/cluster.tf.erb
@@ -142,8 +142,8 @@ resource "libvirt_volume" "admin" {
   base_volume_id = "${libvirt_volume.img.id}"
 }
 
-resource "libvirt_cloudinit" "admin" {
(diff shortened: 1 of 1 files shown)


MalloZup:
@tdaines42 @FedericoCeratto so i backported this to 30-releases.

Feel free to point me doc/wiki for stuff i would need to update on jenkins side :smile:

tdaines42:
So far attempting to get the 42.3 workers to work has not worked. The nodes are deployed but something is causing cloud-init to fail. Also I noticed that 2.1 is using the same workers as 3.0 so we would need to backport this to there as well or make some other changes.

Another option, which I think might be the better one is to backport the label changes to 3.0 so it uses the Leap 15 workers.

tdaines42:
The image is on both prvcld and engcloud the name is jenkins-worker-image.x86_64-1.1.1-Build37.1

PR #525, opened 2018-11-02
WIP: Re-Enable parallel terraform resources

MalloZup:
We should allow TF resources  to be run in parallel.

in the past there were a bug on libvirt side, but the patch got applied.

See here for more details etc..
https://github.com/dmacvicar/terraform-provider-libvirt/issues/402#issuecomment-419500064
(especially this fix `util: don't check for parallel iteration in hash-related functions (rhbz#1581364)`
What we need to check is if the Worker on CI have this patch 

# TODOs:

check if worker have the `libvirt-patch` applied ```x util: don't check for parallel iteration in hash-related functions (rhbz#1581364)```

### DIFF
--- a/caasp-kvm/README.md
+++ b/caasp-kvm/README.md
@@ -98,7 +98,6 @@ sudo virsh pool-start default
 
       * Common options
 
-        -p|--parallelism            Set terraform parallelism (Default: CAASP_PARALLELISM)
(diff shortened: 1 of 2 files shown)


MalloZup:
closing fo the moment as tech-debt, i will try to work on it on free-time and open a PR when i have something ready no need to `park` here

PR #528, opened 2018-11-05
Update requirements

FedericoCeratto:
Fix security issues


MalloZup:
if ci is ok we can merge it. i don't think we should have any issue with that 2

PR #531, opened 2018-11-07
Add sonobuoy-conformance.yaml after Kubernetes upgrade to v1.10.8 (bsc#1114858)

mjura:


k8s-e2e-tests/sonobuoy-conformance.yaml
```diff
@@ -0,0 +1,174 @@
+
```
MalloZup: we might add the commit-hash version of this file

PR #532, opened 2018-11-07
Increate timeout for e2e tests to 14400 sec (bsc#1114858)

mjura:


k8s-e2e-tests/e2e-tests
```diff
@@ -52,10 +52,10 @@ done
 run_tests() {
     rm -f sonobuoy-sonobuoy-conformance.yaml && wget https://raw.githubusercontent.com/cncf/k8s-conformance/bfee9e249bd020314a4989287368134286db8d2e/sonobuoy-conformance.yaml
     # Increase timeout. The default timeout is not enought when running in our test cluster
-    sed -e "s/\"timeoutseconds\": 5400/\"timeoutseconds\": 10800/g" -i sonobuoy-conformance.yaml
+    sed -e "s/\"timeoutseconds\": 5400/\"timeoutseconds\": 14400/g" -i sonobuoy-conformance.yaml
     KUBECONFIG=$KUBECONFIG kubectl apply -f sonobuoy-conformance.yaml
     while ! KUBECONFIG=$KUBECONFIG kubectl logs -n sonobuoy sonobuoy | grep "sonobuoy is now blocking"; do
-        sleep 1
+        sleep 120
```
MalloZup: why do we need here `120` in a while loop?

MalloZup: i think `5` or `3` sec is ok for waiting dinanmically, `120` is to much

mjura: Because it is better than 1 sec of waiting and repeating check.

MalloZup: i think that 1 sec can be to low but imho having `120`  sec can be to much afaik.

If @MaximilianMeister and @tdaines42  @FedericoCeratto  are ok with `120` sec we can merge

MaximilianMeister: > ( if we wait 120 sec in a loop the chances are higher that we need to wait to much as expected)

@MalloZup you're right, however the tests take more than 1.5h so this 2 minutes don't have too much more impact

PR #541, opened 2018-11-14
Improve cleanup

dannysauer:
* Clean up the grep command that purges volumes (and the subseqent xargs) for a bit better readability (and slightly better performance, though that doesn't really matter here)
* Add "suse caas platform" volumes to volume cleanup, which cause terraform to trip if they're not cleaned up
* Add a `-v` to the `rm` commands so what's being deleted is displayed
* Adjust download directory cleanup logic to keep most recent download of each major release image instead of just lexographically "largest"


caasp-kvm/tools/cleanup.sh
```diff
@@ -26,17 +26,49 @@ done
...
+shopt -s extglob
+typeset -A newest=()
+typeset -A keepers=()
+for F in kvm-+([[:alnum:]]); do
+  if [[ -L "$F" ]]; then
+    T="$(readlink -fs "$F")"
+    [[ -f "$T" ]] && keepers[$(basename $T)]=1 
+  fi
+done
+for F in *.qcow2; do
```
MalloZup: can you do a bash function here?  and englobe all this calls, at lest it will more readable

MalloZup:
@dannysauer  after that feel free to merge

caasp-kvm/tools/cleanup.sh
dannysauer: Are you asking to pull the for loop contents out into a function?  I'm not sure that alone would improve readability.  Because bash doesn't allow function prototypes, the whole function definition would need to be before it's called, so all that would happen is that the loop bodies would be indented another level and be shifted up a little higher in the file.

What probably would help is to reorganize the file a bit.  The way I've previously done that is to write a main function at the top, then call main from the end of the file.  Then, by the time "main" is called, bash has parsed the whole file, so function definition location can be done based on readability.  That'll be a bigger change to the script, though; essentially all lines will change.

How do you feel about merging this (assuming proper function) to keep track of "this is when the image cleanup logic was changed" in the git history, and then I'll open a subsequent PR to reorganize the script without altering functionality so it'll all be easier to read?

MalloZup: ok i am fine with this :+1:

MalloZup:
feel free to merge once green :panda_face:

PR #544, opened 2018-11-16
fix missing sudo in velum-interations setup

lcavajani:
I was running some CI but the velum-interations step fails because a sudo is missing:

>>> [velum-bootstrap] Installing Velum Interaction Requirements
grep: /boot/grub2/grub.cfg: Permission denied

### DIFF
--- a/velum-bootstrap/velum-interactions
+++ b/velum-bootstrap/velum-interactions
@@ -76,7 +76,7 @@ check_file() { if [ ! -f $1 ]; then error "File $1 doesn't exist!"; fi }
 # Test methods
 is_ci()
 {
-    if [ -e /boot/grub2/grub.cfg ] && sudo grep 'jenkins-worker' /boot/grub2/grub.cfg > /dev/null; then
+    if [ -f /boot/grub2/grub.cfg ] && sudo grep -q 'jenkins-worker' /boot/grub2/grub.cfg; then
         echo "This image is generated for Jenkins CI purposes."
         return
     fi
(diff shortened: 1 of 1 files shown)


velum-bootstrap/velum-interactions
```diff
@@ -76,7 +76,7 @@ check_file() { if [ ! -f $1 ]; then error "File $1 doesn't exist!"; fi }
 # Test methods
 is_ci()
 {
-    if [ -e /boot/grub2/grub.cfg ] && grep 'jenkins-worker' /boot/grub2/grub.cfg > /dev/null; then
+    if [ -e /boot/grub2/grub.cfg ] && sudo grep 'jenkins-worker' /boot/grub2/grub.cfg > /dev/null; then
```
dannysauer: It's generally preferable to use `sudo -n` inside automation scripts to suppress any prompting due to a misconfigured sudoers; it'd be better for sudo to fail in that case than to hang waiting for input.  If this is run interactively by a user, then sudo failing to open the file will have the same result as no match, so the outcome is unchanged.

Aside from that, it's a tad outside the scope of the change, but I'm a much bigger fan of `grep -q` instead of redirecting output to /dev/null.  It'd probably also be ideal to use `-f` instead of `-e` as well; if that's a block-special or directory, things will likely not go well for grep.  But those are really personal preferences to consider while adding the `-n`. :D

lcavajani: I agree for the -n when this is only used in automation but velum-interactions is sometimes used outside of our CI by other users so they would need to input their password

PR #545, opened 2018-11-19
Set disk caching policy to "unsafe"

FedericoCeratto:
The unsafe policy prevent fsync from KVM VMs to be executed by the hypervisor as well. It seems to be effective as a workaround to unexpectedly high disk write latencies.
It can lead to data loss in the VM in case of power failure on the hypervisor but this is not a concern for CI testbeds.

### DIFF
--- a/caasp-kvm/cluster.tf.erb
+++ b/caasp-kvm/cluster.tf.erb
@@ -181,6 +181,11 @@ resource "libvirt_domain" "admin" {
   }
 <% end -%>
 
+  # used to add disk -> cache="unsafe"
+  xml {
+    xslt = "${file("disk_caching.xsl")}"
+  }
+
   disk {
     volume_id = "${libvirt_volume.admin.id}"
   }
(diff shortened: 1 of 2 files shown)


dannysauer:
This failed the CI because `xml` isn't a top-level key valid in `cluster.tf`.  Looking at the examples in https://github.com/dmacvicar/terraform-provider-libvirt/tree/master/examples/xslt, I suspect that you actually want to move the xml import block into each of the storage resource declarations for admin/master/worker.  Those look like `resource "libvirt_volume" "master"`.

MalloZup:
@FedericoCeratto  it might be that the workers need the devel-unstable version which is not released yet.
If the jenkins worker are using the release version, this is based on the release one.

stefsuse:
@FedericoCeratto  Why was this PR merged? The build was failing due to a known reason, a dependency on an unstable version of a provider we have been using reliably up until now. Merging resulted in the need for PR #550 , but this should have happened in *this* PR as the changes are dependent.

PR #548, opened 2018-11-20
add script to check velum page status

lcavajani:
This is meant to be used in automation to be sure Velum is running before starting velum-interactions.

This is especially required for hyper-v where the Velum page takes ~45s to be up


misc-tools/check_velum_status
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# USAGE: ./check_velum_status ./environment.json 120
+if [[ ! -f $1 ]];then echo "environment file not found: $1" && exit 1; fi
```
dannysauer: If an `if` block is used, the commands should be on separate lines for readability.  The exit also really doesn't depend on the echo statement returning true. :)  Thank you for using `[[` in the conditional, though. :D

misc-tools/check_velum_status
```diff
@@ -0,0 +1,17 @@
...
+
+ADMIN_IP=$(cat $1 | jq '.minions[0]'.addresses.publicIpv4 -r)
+TIMEOUT=${2:-120}
+VELUM_STATUS=
+
+echo "checking velum status..."
+for elapsed in $(seq 0 $TIMEOUT); do
+  VELUM_STATUS=$(curl -k -I -s -o /dev/null -w "%{http_code}" --connect-timeout 2 https://$ADMIN_IP)
+  [[ $VELUM_STATUS -eq 200 ]] && echo "ready, page returned: $VELUM_STATUS" && exit 0
```
dannysauer: IMO, the exit is the important part here; the success or failure of `echo` is irrelevant.  I do like the inline'd condition and command idiom for one command, and it's admittedly absurdly unlikely that `echo` will return non-zero.  However, when there are two things which should both always happen, I'd suggest either converting to an if block, or grouping the commands together with curly braces, as in
`[[ conditional ]] && { cmd1; cmd2; }`
The braces don't start a subshell like parens do, so allow grouping without introducing the overhead of a fork.

Pick whichever style you prefer, but I think this line and the one on Line 3 should be consistent.

misc-tools/check_velum_status
lcavajani: mouarf, fixed ;)

misc-tools/check_velum_status
dannysauer: Again for the record, I do really like the use of `seq` and early exit mechanism employed here. :)

lcavajani: Thanks for the info, that's very valuable ;)

You're right that I start the assumption that echo will not fail... I have used the "if" method to be consistent.

lcavajani:
There is still no PR but this will come soon

misc-tools/check_velum_status
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
MalloZup: add here after the bash, 
add the use-case of the script and in which context it is supposed to be used. If you use it in a pipeline, feel free to mention the link.
this `This is meant to be used in automation to be sure Velum is running before starting velum-interactions.`

Less context individual saved information in brain of people we have and more we will have a CI without maintainers :sunflower:

lcavajani: just found this https://github.com/kubic-project/automation/blob/master/misc-tools/wait-for-velum :dancer: 

so there's no need for this PR...

MalloZup: @lcavajani ok nice :+1:  you can improve doc on that script and reference the pipe used there if need this could spare time for the next

lcavajani:
I have created the PR for jenkins-library here: https://github.com/kubic-project/jenkins-library/pull/226

PR #552, opened 2018-11-28
cleanup.sh: delete kvm-devel volume

FedericoCeratto:
Also print out some of the resources before deletion.

### DIFF
--- a/caasp-kvm/tools/cleanup.sh
+++ b/caasp-kvm/tools/cleanup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Cleaning up KVM: delete VMs, networks, volumes and terraform states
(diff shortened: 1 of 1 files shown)


caasp-kvm/tools/cleanup.sh
```diff
@@ -69,9 +75,9 @@ done
 shopt -u extglob
 popd  > /dev/null
 
-echo "--> Cleanup screenshots"
+echo "--> Cleaning up screenshots"
 pushd $DIR/../../velum-bootstrap > /dev/null
-rm -vrf screenshots 
+rm -vrf screenshots
```
MalloZup: do we have a space here?

FedericoCeratto: We did

PR #554, opened 2018-12-07
hide the outdated RMT/SMT mirror warning in the devenv

MaximilianMeister:
we'll likely never use a mirror in the devenv


MaximilianMeister:
I just realized i need to put all the exec's under a `if !vanilla` condition, right now it's only on the admin node

PR #555, opened 2018-12-10
add tool to list images on OBS in a download repository

lcavajani:
I want to use this to trigger a build for a platform when a new image is available on OBS.

Use in CI to list the images in a download repository on OBS so we can use the output file to detect a new image available and trigger a build.


misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,94 @@
...
+
+import argparse
+from HTMLParser import HTMLParser
+import json
+import re
+import requests
+
+
+# HTML Parser
+class MyHTMLParser(HTMLParser):
```
MalloZup: rename class to `CIHTMLParser` instead of `my` whatever :smile:

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,94 @@
...
+    files = []
+
+    def handle_starttag(self, tag, attrs):
+        if tag == "a":
+            for attr in attrs:
+                # ('href', 'SUSE-CaaS-Platform-4.0-for-XEN.x86_64.qcow2')
+                self.files.append(attr[1])
+
+
+# Parse command line arguments
```
MalloZup: create a function here

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,94 @@
...
+    page = request.content
+except (requests.ConnectionError, requests.HTTPError, requests.Timeout) as e:
+    print("connection failed: {0}".format(e))
+
+# Parse HTML content of the page
+html_parser = MyHTMLParser()
+html_parser.feed(page)
+result = list(set(html_parser.files))
+
+# Remove unnecessary files
```
MalloZup: this is a function, we could also use `SUSE` as var instead of passing it directly

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,94 @@
...
+        json_output["openstack"].append(f)
+
+    if re.search("vmware", f, re.IGNORECASE):
+        json_output["vmware"].append(f)
+
+    if re.search("xen", f, re.IGNORECASE):
+        if not re.search("kvm", f, re.IGNORECASE):
+            json_output["xen"].append(f)
+
+# Outputs
```
MalloZup: this comment is unneeded. 

I woul also split the code to a function

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if re.search("openstack", f, re.IGNORECASE):
+            json_output["openstack"].append(f)
+
+        if re.search("vmware", f, re.IGNORECASE):
+            json_output["vmware"].append(f)
+
+        if re.search("xen", f, re.IGNORECASE):
+            if not re.search("kvm", f, re.IGNORECASE):
+                json_output["xen"].append(f)
```
dannysauer: This set of repeated code could be reduced a couple of ways.  One, with a regexp storing the match:
```python
platform = re.search("kvm|hyperv|openstack|vmware|(?<!kvm-and-)xen", f, re.IGNORECASE)
if platform is not None:
  json_output[platform.group(0).lower()].append(f)
```

Alternatively, since regexps are scary and these are fixed strings anyway:
```python
platforms = [
    "hyperv",
    "kvm", # must appear before xen in list
    "openstack",
    "vmware",
    "xen",            
]
for f in file_list
    for platform in platforms:
        if platform in f
            json_output[platform].append(f)
            break
```
Either approach allows for more easily handling "nothing matched in file" kind of logs.

dannysauer: I like the regex because I like Perl, but the second approach is probably more readable to people who don't have that specific form of brain damage. :D

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
lcavajani: only python3, I'll keep that in mind for the next time

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,161 @@
...
+    print(json_output)
+    write_file(result_file, json_output)
+
+    if download_checksum:
+        print("INFO: Retrieving sha256 checkfum files\n")
+        if not os.path.exists(checksum_dir):
+            os.makedirs(checksum_dir)
+
+        for f in sha256files:
+            link = "{0}{1}".format(url, f)
```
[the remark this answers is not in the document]
lcavajani: I agree

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,161 @@
...
+    download_checksum = args.download_checksum
+    checksum_dir = args.checksum_dir
+    insecure = args.insecure
+    url = args.url
+
+    if url[-1] is not "/":
+        url = "{0}/".format(url)
+
+    if checksum_dir[-1] is not "/":
+        checksum_dir = "{0}/".format(checksum_dir)
```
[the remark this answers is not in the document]
lcavajani: it's much better with os.path.join() ;)

misc-tools/list_image_repo.py
lcavajani: I went for the first approach, my brain is not as damaged a yours for sure but the regex is not so complicated :smile:

lcavajani:
@dannysauer I used flake but PEP8 says "For one liner docstrings, please keep the closing """ on the same line."

Thanks again for the review, that's very helpful ;)

misc-tools/list_image_repo.py
```diff
@@ -0,0 +1,146 @@
...
+    json_output = {}
+    json_output["url"] = url
+
+    for f in file_list:
+        platform = re.search(platforms_regex, f, re.IGNORECASE)
+        if platform is not None:
+            # Create key
+            if platform.group(0).lower() not in json_output:
+                json_output[platform.group(0).lower()] = []
+            json_output[platform.group(0).lower()].append(f)
```
dannysauer: This is a job for a defaultdict. :)

This exact thing is the first example in the documentation: https://docs.python.org/2/library/collections.html#defaultdict-objects

lcavajani: that's convenient, thanks

PR #564, opened 2018-12-17
jenkins-jobs: job-template: Disable the codenarc test

hwoarang:
The codenarc test is only useful to run Groovy codying style checks
and it's currently failing to due disabled k8s integration so lets
disable it.


hwoarang:
No longer needed since k8s integration has been enabled again

PR #565, opened 2018-12-17
Handle kubectl scale failures

FedericoCeratto:
Add retry mechanism to handle failures currently impacting release 3.0

### DIFF
--- a/k8s-pod-tests/k8s-pod-tests
+++ b/k8s-pod-tests/k8s-pod-tests
@@ -210,7 +210,11 @@ scale_pod_in_steps() {
       new_num=$NUM_REPLICAS
     fi
     echo "Setting replicas to $new_num"
-    kubectl scale --replicas=$new_num --kubeconfig="$KUBECONFIG_FN" deployment/$NAME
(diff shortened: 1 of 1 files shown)


k8s-pod-tests/k8s-pod-tests
```diff
@@ -210,7 +210,11 @@ scale_pod_in_steps() {
       new_num=$NUM_REPLICAS
     fi
     echo "Setting replicas to $new_num"
-    kubectl scale --replicas=$new_num --kubeconfig="$KUBECONFIG_FN" deployment/$NAME
+    for retries in $(seq 1 10); do
+      kubectl scale --replicas=$new_num --kubeconfig="$KUBECONFIG_FN" deployment/$NAME && break
+      echo "[$retries / 10] kubectl scale failed. Retrying in 10 seconds..."
+      sleep 10
```
stefsuse: 10 seconds is a very long time to wait, and if this error is not recoverable this change will just guarantee that CI will fail 2 minutes later than it should have.

PR #569, opened 2018-12-20
Fixes to how container tests are handled

tdaines42:
Fixes for differences between how macOS and leap 15 handle docker images. Also added dockerignore for python cache files.

### DIFF
--- a/tests/util/container_info.py
+++ b/tests/util/container_info.py
@@ -1,0 +1,10 @@
+def network_names(container):
+    return list(container.attrs['NetworkSettings']['Networks'])
+
+
+def get_network_by_name(container, name, partial=False):
+    for network_name in network_names(container):
+        if name == network_name:
+            return network_name
+        elif partial and name in network_name:
+            return network_name
--- a/tests/.dockerignore
+++ b/tests/.dockerignore
@@ -1,0 +1,2 @@
+**/__pycache__/
+**/.pytest_cache/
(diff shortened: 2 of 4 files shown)


tests/util/container_info.py
```diff
@@ -0,0 +1,10 @@
+def network_names(container):
+    return list(container.attrs['NetworkSettings']['Networks'])
+
+
+def get_network_by_name(container, name, partial=False):
+    for network_name in network_names(container):
```
dannysauer: Should `name` and `network_name`  be forced to lower-case to make the match case-insensitive?

I'm also wondering what happens if name is both an exact match for a network and a substring of another network.  If `partial` is set to true, the returned value then depends on the ordering of the elements returned by network_names, and I'm not sure that's deterministic.  Perhaps sorting the network_names first by length and then lexographically would be worthwhile, in the very slim chance that becomes relevant?  Or perhaps that's overkill. :)

tdaines42: My idea with this is that the person using this will know which networks are going to be attached to the container. The reason for `partial` was due to how that network names were formed differently on macOS vs linux. On macOS network name was allowed to be spinal case where on Leap it was just all squashed together.

tdaines42: I could add some of the other things like making it case insensitive and sorting but I'm not sure if it matters at this point.

dannysauer: Yeah - both are probably overkill.  Maybe "in the future"...

It's also underscores w/ the Docker I'm using on Tumbleweed, BTW.

PR #572, opened 2018-12-27
README.md: Update script name

jvanz:
Updates some references within README.md file to a script not present in the repository, the caasp-devenv. Now uses caasp-kvm instead.


jvanz:
Invalid change. The script mentioned in the commit is in the root of the repository. I didn't notice that. Sorry

Issue #577, opened 2019-01-04
Wrong bundler version


tdaines42:
A recent update to Bundler seems to be causing the install to fail in the velum devel container.

This results in the following error while the caasp velum devel image is being built:
```
ERROR:  Error installing bundler:
bundler requires Ruby version >= 2.3.0
```
Most likely this line `gem install --no-ri --no-rdoc bundler -n /usr/bin && \` in `caasp-kvm/tools/build-velum-image::build_fresh_image()` needs to be updated to something like `gem install --no-ri --no-rdoc bundler -n /usr/bin -v 1.17.3 && \`. 

I tested this manually and the image seemed to build just fine.

suseclee:
Ty. This is very good analogy and exact solution. This shows how a simple upstream change can end up with building failure.

stefsuse:
Closing now that backport is done.

PR #578, opened 2019-01-05
Fix Issue #577: ruby bundler update

suseclee:
## What does this PR change?

After bundler was updated to 2.0.0 from rubygems.org in 01/03/2019, velum-development:0.0, which has ruby2.1 failed to install ruby bundler because new ruby bundler requires >= ruby2.3.0 and gem installation uses highest version if not specified.
After it failed installing bundler, it does not install any dependent gems on the velum image template.
This causes to fail to deploy static pods in admin node. And cluster build fails with "Waiting for Velum to initialize". So simple solution is locking the safe version on bundler.

### DIFF
--- a/caasp-kvm/tools/build-velum-image
+++ b/caasp-kvm/tools/build-velum-image
@@ -90,7 +90,7 @@ build_fresh_image() {
                     bash -c "
                       cd /srv/velum-latest && \
                       mkdir -p /var/lib/velum && \
-                      gem install --no-ri --no-rdoc bundler -n /usr/bin && \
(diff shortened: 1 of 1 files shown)


caasp-kvm/tools/build-velum-image
```diff
@@ -90,7 +90,7 @@ build_fresh_image() {
                     bash -c "
                       cd /srv/velum-latest && \
                       mkdir -p /var/lib/velum && \
-                      gem install --no-ri --no-rdoc bundler -n /usr/bin && \
+                      gem install --no-ri --no-rdoc bundler -v 1.17.3 -n /usr/bin && \
```
tdaines42: LGTM

PR #583, opened 2019-01-08
[caasp-openstack-terraform] Remove deploy script in favor of direct terraform cli

lcavajani:
## What does this PR change?

This PR does:
* Remove `caasp-openstack` in order to use only `terraform` cli
* Rename `identifier` variable to `stack_name` which is closer to what we use in automation
* Remove region variable in TF file as this is get by environment variable

I have tested this on ECP. This script is currently not used in the automation, `caasp-openstack-heat` is used.

So basically, instead of running **./caasp-openstack apply**, we can use **terraform apply** and a lot of other commands.

Why removing `caasp-openstack` script ?

In order to use the TF files, it is necessary to use `caasp-openstack` script. This script does the following:

1. Read OpenStack environment variables `$OS_*`
2. Create `ssh/` dir in current directory
3. Create RSA key in `ssh/id_caasp`
4. Run terraform with the available commands `apply,refresh,plan,destroy,output,show`
5. Run terraform with parameters `-var-file=openstack.tfvars`
6. Show help

Here are the drawbacks/limitations of each point:

1. This is already done by terraform so there is no need to do it before running terraform
2. This can be done by terraform
3. This can be done by terraform
4. **This limits us to a bunch of commands, we can't add extra parameters for the available commands** (e.g -var 'stack_name=yellow')
5. Terraform can read by default this file if it is named `terraform.tfvars`
6. Using `terraform help` provides the same kind of help


lcavajani:
I've fixed a remaining debug code

PR #587, opened 2019-01-17
Add testrunner script

FedericoCeratto:
Initial version of a script to replace jenkins-library. It's meant to be run on http://jenkins.caasp.suse.net initially, later on on ci.suse.de and developer workstations.
It supports KVM and Bare-metal. OpenStack and complex pipelines like upgrade testing will be implemented on it.


misc-tools/test_kubic/testrunner
```diff
@@ -0,0 +1,21 @@
...
+
+# TODO: allow venv dir override
+${SDIR}/../setup-python timeout-decorator
+
+# Please track what components require each lib!
+
+# wait-for-velum
+${SDIR}/../setup-python requests
+
+export VENVDIR=${WORKSPACE:-~}/py3venv
```
hwoarang: can we change the default option to something less dangerous like `$SDIR` or `/tmp` for example?

hwoarang: nit: Should the test_kubic.py file has the +x bit set?

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,678 @@
...
+    conf.master_ram = "4096"
+    conf.worker_cpu = "4"
+    conf.worker_ram = "4096"
+    conf.vanilla = False
+    conf.netlocation = "provo"
+    conf.channel = "devel"
+    conf.replica_count = "5"
+    conf.replicas_creation_interval_seconds = "5"
+    conf.podname = "default"
+    conf.image = replace_vars("file://${WORKSPACE}/automation/downloads/kvm-devel")
```
hwoarang: Unfortunately not all of these options are self-explanatory. What is `netlocation` for? what is `change_author` for and why would I need to tweak that?

misc-tools/test_kubic/testrunner
hwoarang: I also think that adding some docstrings to all these methods is necessary :)

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,678 @@
...
+
+@timeout(25)
+@step()
+def clone_repo(gitBase="", branch="", ignorePullRequest="", repo=""):
+    sh("git clone --depth 1 {}/{}".format(gitBase, repo))
+
+@step()
+def retrieve_code():
+    """Retrieve Code"""
+    gitBase = "https://github.com/kubic-project"
```
hwoarang: should this be a parameter to allow people to clone from different locations?

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,678 @@
...
+    branch = getvar('CHANGE_TARGET')
+    ignorePullRequest = False
+    # TODO: parallel
+    clone_repo(gitBase=gitBase, branch=branch, ignorePullRequest=ignorePullRequest, repo="automation")
+    clone_repo(gitBase=gitBase, branch=branch, ignorePullRequest=ignorePullRequest, repo="salt")
+    clone_repo(gitBase=gitBase, branch=branch, ignorePullRequest=ignorePullRequest, repo="velum")
+    clone_repo(gitBase=gitBase, branch=branch, ignorePullRequest=ignorePullRequest, repo="caasp-container-manifests")
+    clone_repo(gitBase=gitBase, branch=branch, ignorePullRequest=ignorePullRequest, repo="caasp-services")
```
hwoarang: should you also provide a way for people to clone additional repos?

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,690 @@
...
+    """Cleanup"""
+    sh('rm -rf ${WORKSPACE} || : ')
+    create_workspace_dir()
+    sh('mkdir -p ${WORKSPACE}/logs')
+    sh('chmod a+x ${WORKSPACE}')
+    if conf.stack_type == 'bare-metal':
+        return
+
+    sh('virsh net-undefine caasp-dev-net || : ')
+    sh('virsh net-destroy caasp-dev-net || : ')
```
MaximilianMeister: is that network name even still used?

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,690 @@
...
+    sh('virsh net-undefine net || : ')
+    sh('virsh net-destroy net || : ')
+    sh('for i in $(virsh list --all --name);do echo $i;virsh destroy $i || : ;done')
+    sh('for i in $(virsh list --all --name);do echo $i;virsh undefine $i;done')
+    sh('for fn in $(virsh vol-list default|awk \'/var/ {print $2}\'); do echo $fn; virsh vol-delete $fn ; done')
+    sh('virsh list --all')
+    sh('virsh net-list --all')
+    sh('virsh pool-list --all')
+    sh('virsh vol-list default')
```
MaximilianMeister: as we use quite a lot of `virsh` commands, maybe it makes sense to create a central function with commands/actions as args?

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,690 @@
...
+    conf.worker_ram = "4096"
+    conf.vanilla = False
+    conf.netlocation = "provo"
+    conf.channel = "devel"
+    conf.replica_count = "5"
+    conf.replicas_creation_interval_seconds = "5"
+    conf.podname = "default"
+    conf.image = replace_vars("file://${WORKSPACE}/automation/downloads/kvm-devel")
+
+    if '-h' in sys.argv:
```
MaximilianMeister: i'd add a `--help` here too. i first tried with that, and then gave up

FedericoCeratto: Done

misc-tools/test_kubic/test_kubic.py
FedericoCeratto: This is an initial version of the script, we'll add more complex features as we go.

misc-tools/test_kubic/test_kubic.py
FedericoCeratto: I don't see a benefit in wrapping virsh in terms of readability. The entire script is meant to be as explicit as possible so that developers can tweak it or copy-paste bash commands to test them.

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,690 @@
...
+    sh('virsh net-destroy net || : ')
+    sh('for i in $(virsh list --all --name);do echo $i;virsh destroy $i || : ;done')
+    sh('for i in $(virsh list --all --name);do echo $i;virsh undefine $i;done')
+    sh('for fn in $(virsh vol-list default|awk \'/var/ {print $2}\'); do echo $fn; virsh vol-delete $fn ; done')
+    sh('virsh list --all')
+    sh('virsh net-list --all')
+    sh('virsh pool-list --all')
+    sh('virsh vol-list default')
+    sh('docker rm -f $(docker ps -a -q) || :')
+    sh('docker system prune --all --force --volumes || :')
```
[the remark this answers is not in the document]
MaximilianMeister: we can adapt that later

misc-tools/test_kubic/testrunner
FedericoCeratto: (discussed on Rocket Chat)

misc-tools/test_kubic/pipeline.example.groovy
```diff
@@ -0,0 +1,24 @@
+
+// Run on jenkins.caasp.suse.net, use KVM
+
+pipeline {
+    agent any
+    stages {
+        stage('Run') {
```
hwoarang: OK so the entire script is run as a single pipeline step correct? That may make debugging somewhat difficult?

I am guessing it's not possible to have multiple `stages` with this script like we have right now in our pipelines?

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,690 @@
...
+    sh('virsh net-undefine net || : ')
+    sh('virsh net-destroy net || : ')
+    sh('for i in $(virsh list --all --name);do echo $i;virsh destroy $i || : ;done')
+    sh('for i in $(virsh list --all --name);do echo $i;virsh undefine $i;done')
+    sh('for fn in $(virsh vol-list default|awk \'/var/ {print $2}\'); do echo $fn; virsh vol-delete $fn ; done')
+    sh('virsh list --all')
+    sh('virsh net-list --all')
+    sh('virsh pool-list --all')
+    sh('virsh vol-list default')
+    sh('docker rm -f $(docker ps -a -q) || :')
```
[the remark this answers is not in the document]
FedericoCeratto: As discussed on RC we had many issues with "leftover" VMs, images etc. The current implementation is meant to be used on bare metal hypervisors that do one CI run at a time. If you can provide a precise pruning command I'll be happy to add it.

misc-tools/test_kubic/test_kubic.py
```diff
@@ -0,0 +1,772 @@
...
+        cmd = "tox -e {role}-{status} -- --hosts {fqdn} --junit-xml" \
+           " testinfra-{role}-{index}.xml -v".format(**minion)
+        shp("${WORKSPACE}/automation/testinfra", cmd, env=env)
+
+    #junit "testinfra-${minion.role}-${minion.index}.xml"
+
+
+
+@timeout(600)
+@step()
```
MalloZup: would be nice imho if we could split the steps in a `steps.py` file and the `main.py` scheduling,
Imho we could create a dir with a Readme.md some doc, and the python components.

PR #592, opened 2019-01-28
velum-bootstrap: check content disposition instead of url

vitoravelino:
The spec for checking the kubeconfig download was previously done
through checking an url associated with the automatic download.

A better way of checking the autoamtic download is to check the
content disposition in the response header.

### DIFF
--- a/velum-bootstrap/spec/features/03-download-kubeconfig.rb
+++ b/velum-bootstrap/spec/features/03-download-kubeconfig.rb
@@ -43,17 +43,17 @@ feature "Download Kubeconfig" do
     expect(page).to have_text("You will see a download dialog")
     puts "<<< User is redirected back to velum"
 
+    puts ">>> User sees a download button to download file if neeeded"
+    expect(page).to have_text("Click here if the download has not started automatically.")
+    puts "<<< User sees a download button to download file if neeeded"
+
     puts ">>> User is prompted to download the kubeconfig"
-    download_uri = URI(page.html.match(/window\.location\.href = "(.*?)"/).captures[0])
-
-    http = Net::HTTP.new(download_uri.host, download_uri.port)
-    http.use_ssl = true
-    http.verify_mode = OpenSSL::SSL::VERIFY_NONE
-
-    request = Net::HTTP::Get.new(download_uri.request_uri)
-    response = http.request(request)
-
-    File.write("kubeconfig", response.body)
+    content_disposition = page.response_headers["Content-Disposition"]
+    expect(content_disposition).to eq("attachment; filename=kubeconfig")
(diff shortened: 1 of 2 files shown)


vitoravelino:
I tried to fix some rubocop complains but that should go in a dedicated PR instead of this one.

MalloZup:
@vitoravelino we can fix rubocop later no worry.

Or we can autogenerate the stuff and blacklist

vitoravelino:
> @vitoravelino imho the `kubeconfig` file is used later, so we still need to dowload it, check out the failing test

Indeed. I'll take a look at it.

vitoravelino:
Finally green! :tada:

velum-bootstrap/spec/features/03-download-kubeconfig.rb
```diff
@@ -43,17 +43,17 @@
...
-
-    http = Net::HTTP.new(download_uri.host, download_uri.port)
-    http.use_ssl = true
-    http.verify_mode = OpenSSL::SSL::VERIFY_NONE
-
-    request = Net::HTTP::Get.new(download_uri.request_uri)
-    response = http.request(request)
+    puts ">>> User sees a download button to download file if neeeded"
+    expect(page).to have_text("Click here if the download has not started automatically.")
+    puts "<<< User sees a download button to download file if neeeded"
```
hwoarang: is this supposed to duplicate line 46?

velum-bootstrap/spec/features/03-download-kubeconfig.rb
```diff
@@ -43,17 +43,17 @@
...
 
-    File.write("kubeconfig", response.body)
+    puts ">>> User is prompted to download the kubeconfig"
+    content_disposition = page.response_headers["Content-Disposition"]
+    expect(content_disposition).to eq("attachment; filename=kubeconfig")
     puts "<<< User is prompted to download the kubeconfig"
```
hwoarang: is this supposed to duplicate line 50?

velum-bootstrap/spec/features/03-download-kubeconfig.rb
vitoravelino: It's not duplicated, one starts with `>>>` and the other with `<<<`. I guess the pattern was to indicate that the test passed through that stage.

hwoarang: Oh yes you are right my bad.

velum-bootstrap/spec/features/03-download-kubeconfig.rb
hwoarang: nm

PR #595, opened 2019-01-29
Update rubocop blacklist

MalloZup:
## What does this PR change?
Update rubocop blacklist so we have green in travis

ro


hwoarang:
I understand this update is needed to make tests happy but can you explain why tests were failing before? Is the blacklist the right thing to do and if yes why? all that should go to the commit message :)

MalloZup:
@ereslibre feel free to ping me once you have something

PR #596, opened 2019-01-29
Add VMware support

FedericoCeratto:
And a little cleanup.

### DIFF
--- a/misc-tools/download-image
+++ b/misc-tools/download-image
@@ -12,39 +12,56 @@ import sys
 import urlparse
 from HTMLParser import HTMLParser
 
-KVM_REGEX = '.*KVM.*x86_64-.*\.qcow2$'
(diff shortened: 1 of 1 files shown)


misc-tools/download-image
```diff
@@ -254,7 +272,7 @@ def get_channel_url(args):
 
     if args.type == "docker":
         base_url += 'images_container_base'
-    elif args.type in ["kvm", "openstack"]:
+    elif args.type in ["kvm", "openstack", "vmware"]:
```
MalloZup: `nit` we might put this maybe in a `tuple` where we define at the TOP of the file.

MalloZup: but this is ok, could be a later refactoring idea for freezing the list in immutable data

PR #547, opened 2018-11-20
Make SLE15 based images the default.

bergmannf:
This will make the `SLE15` based environment the default.

For now `release` and `staging_*` automation will assume `sle12` as underlying image.
Once they have been moved to `sle15` most of the conditional introduced can be removed.

PR #598, opened 2019-01-30
misc-files: download-urls: Switch to SLE15 images

hwoarang:
We should be using the SLE15 based images for development.


MalloZup:
we have other PRs targeting SLE15. (really old). pleasy depracted them/close them, or discuss with @bergmannf if we need them

hwoarang:
Closing in favor of #547 547

PR #599, opened 2019-01-30
.github: Update PR template

hwoarang:
- Remove branch-2.1 references
- Add commit message guidelines
- Add more context to the PR and documentation section
- Fix formatting for issue tracking

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,17 +1,29 @@
 ## What does this PR change?
+- What was wrong with the existing code?
+- How does this PR fixes it?
 
-**add description**
-(Previous behaviour, Current behaviour).
+Please be as verbose as possible. Remember others need time
+to context switch and review your PR, so the more information
+you provide to them the more chances you have to see your
+PR merged.
+
+## Commit message
+Does your commit message follow the usual Git guidelines?
+
+- Separate subject from body with blank line
+- Limit the subject to 50 characters if possible
+- Commit body should be line wrapped at 75 columns
+- Use the body to explain what has changed, why it was
+  changed and how this change makes things better.
 
 ## Backport to other branches
 Use  `git cherry-pick` backport a commit from master to branches
 
-- [ ] release-2.1 
 - [ ] release-3.0
 
 ## Documentation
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,29 @@
 ## What does this PR change?
+- What was wrong with the existing code?
```
MalloZup: i'm not a fun of this 2 first question imho we could omit them

hwoarang: You can omit that it's just a guideline. Sometimes it makes sense, some other times it doesn't

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,29 @@
...
+to context switch and review your PR, so the more information
+you provide to them the more chances you have to see your
+PR merged.
+
+## Commit message
+Does your commit message follow the usual Git guidelines?
+
+- Separate subject from body with blank line
+- Limit the subject to 50 characters if possible
+- Wrap the body at 72 characters
```
ereslibre: :thinking:

.github/PULL_REQUEST_TEMPLATE.md
MalloZup: @hwoarang  from highlevel pov imho i thinkk that this 2 statements are better then nothing, but we omit the case where someone add a new feature, or nothing is wrong with the code, this case. but yep , we can have it as it is. imho

.github/PULL_REQUEST_TEMPLATE.md
MalloZup: maybe words?

ereslibre: I hate to say this, but if we want this kind of guidelines to be consistent we probably need a linting job for the commit message itself, otherwise it's a best effort usually ignored.

ereslibre: (And the linting as warning, because you could link to a long URL that you should not wrap or something similar)

MalloZup: i think we can merge it , it is an improvement already

PR #603, opened 2019-02-01
[WIP]  Use the same conformance strategy as in master

ereslibre:
Instead of doing conformance in a complete different way for 3.0,
let's do the same strategy as in master. This will probably need
to be paramaterized because different sonobuoy versions support
different cluster versions and the drift of versions between `master`
and `release-N.0` branches could force us to use different sonobuoy
versions in those branches.

In any case, apart from the sonobuoy version to be used we should
be consistent in the way that we use sonobuoy as a whole.


ereslibre:
Closing because the current sonobuoy package version baked into the CI image is incompatible with K8s 1.10 used by release 3.0.

PR #604, opened 2019-02-01
Allow to specify a specific sonobuoy repository and version

ereslibre:
The script `e2e-tests` will now accept parameters `--sonobuoy-image`
and `--sonobuoy-version` so it's possible to run the conformance tests
specifically for the specified sonobuoy versions.

This will typically come as a jenkins pipeline parameter, and allows us
to not modify the status of the worker other than creating containers.

Depends on: https://github.com/kubic-project/jenkins-library/pull/248

### DIFF
--- a/k8s-e2e-tests/e2e-tests
+++ b/k8s-e2e-tests/e2e-tests
@@ -5,6 +5,8 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 ARTIFACTS_PATH=$DIR/results
 KUBECONFIG=
 SONOBUOY_RUN_ARGS=()
+SONOBUOY_IMAGE="gcr.io/heptio-images/sonobuoy"
+SONOBUOY_VERSION="latest"
 E2E_WAIT_TIME=${E2E_WAIT_TIME:-120} # How long to wait for the test to finish in minutes
 
 USAGE=$(cat <<USAGE
@@ -17,6 +19,8 @@ Usage:
 Other:
     --e2e-focus              set the e2e tests to focus on
     --e2e-skip               set the e2e tests to skip
+    --sonobuoy-image         set the sonobuoy image to be used ('gcr.io/heptio-images/sonobuoy' by default)
+    --sonobuoy-version       set the sonobuoy version to be used ('latest' by default)
     --artifacts <DIR>        directory where junit XML files are stored
 
 USAGE
@@ -37,7 +41,6 @@ while [[ $# > 0 ]] ; do
     -k|--kubeconfig)
       KUBECONFIG="$(realpath $2)"
       check_file $KUBECONFIG
-      SONOBUOY_RUN_ARGS+=(--kubeconfig $KUBECONFIG)
       shift
       ;;
     --e2e-focus)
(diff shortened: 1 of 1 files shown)


hwoarang:
I like keeping the host clean but would it make sense to give people the option to use the one installed of the host if needed?

ereslibre:
> I like keeping the host clean but would it make sense to give people the option to use the one installed of the host if needed?

IMO we should remove `sonobuoy` from the base image of the CI, so we relay to any specific version if we want, even without the need to package; we should be able to run all kind of pipelines in any base system (dumb runner).

hwoarang:
I will let you debate this with @MaximilianMeister  :)

k8s-e2e-tests/e2e-tests
```diff
@@ -60,28 +72,19 @@ while [[ $# > 0 ]] ; do
...
 }
 
 run_tests() {
-    if ! command -v sonobuoy > /dev/null; then
-        abort "Package sonobuoy needs to be installed"
-    fi
-
-    test_config
-
     echo "${SONOBUOY_RUN_ARGS[@]}" | xargs -t sonobuoy run
```
MaximilianMeister: i think here it will not use the function, but the `sonobuoy` in $PATH

ereslibre: Thanks! Overlooked it :)

ereslibre:
I got a green pipeline here, but it failed with a strange error when reporting the success to github :)

k8s-e2e-tests/e2e-tests
```diff
@@ -60,52 +71,41 @@ while [[ $# > 0 ]] ; do
...
+    if sonobuoy status | grep "Sonobuoy has completed"; then
         # Create the artifacts path
         mkdir -p $ARTIFACTS_PATH
         # Copy results from the container
-        sonobuoy retrieve --kubeconfig $KUBECONFIG $ARTIFACTS_PATH
+        sonobuoy retrieve $ARTIFACTS_PATH
         # Extract conformance tests tarball results
         tar -xzf ${ARTIFACTS_PATH}/*_sonobuoy_*.tar.gz -C ${ARTIFACTS_PATH}/
-    elif sonobuoy status --kubeconfig $KUBECONFIG 2>&1 | grep "Sonobuoy is still running"; then
+    elif sonobuoy status | grep "Sonobuoy is still running"; then
```
MalloZup: This might be not really safe to rely on, e.g if upstram Change that string we fail. Is there a more pragmatic way for it? Maybe we should add hack here comment and push upstream a fix for having a code or something more computer like 🦄

MalloZup: Imho We might check save the Pid is more Unix then grepping golang hipster strings🎶😎😺😁

ereslibre: I think we can improve in a later iteration, it's a bit more intrincated than saving the PID :), since these tests run on the cluster and the result is within the cluster. We can do something better still, yes :)

hwoarang:
Build error is unrelated so lets get it in

k8s-e2e-tests/e2e-tests
MalloZup: Yop agree, I just spot it. We can merge this

PR #606, opened 2019-02-04
Jenkinsfile.codenarc: Use docker cmdline directly instead of k8s Jenkins plugin

hwoarang:
The codenarc pipeline is the only one which depends on the k8s Jenkins
plugin. As such, we can use docker directly in order to be able to drop
the plugin from the server.

## What does this PR change?
- Unnecessary dependency to k8s JK plugin
- Uses docker cmd directly

## Backport to other branches
Use  `git cherry-pick` backport a commit from master to branches

- [ X] release-3.0

This will need backport to release-3.0 branch once it's green

### DIFF
--- a/Jenkinsfile.codenarc
+++ b/Jenkinsfile.codenarc
@@ -10,34 +10,18 @@ githubCollaboratorCheck(
     user: env.CHANGE_AUTHOR,
     credentialsId: 'github-token')
 
-def label = "salt-automation-${UUID.randomUUID().toString()}"
+node("leap15.0&&caasp-pr-worker") {
+    stage('Retrieve Code') {
+         checkout scm
+    }
 
-podTemplate(label: label, containers: [
-        containerTemplate(
-            name: 'codenarc',
-            image: 'registry.suse.de/devel/casp/ci/opensuse_leap_42.3_containers/jenkins-codenarc-container:latest',
-            alwaysPullImage: true,
-            ttyEnabled: true,
-            command: 'cat',
-            envVars: [
-                envVar(key: 'http_proxy', value: env.http_proxy),
-                envVar(key: 'https_proxy', value: env.http_proxy),
-            ],
-        ),
-]) {
-    node(label) {
-        stage('Retrieve Code') {
-            checkout scm
-        }
-
+    docker.image('registry.suse.de/devel/casp/ci/opensuse_leap_42.3_containers/jenkins-codenarc-container:latest').inside('-v ${WORKSPACE}:/codenarc') {
         stage('Style Checks') {
-            container('codenarc') {
-                try {
-                    sh 'java -classpath /groovy-all-2.4.6.jar:/CodeNarc-1.1.jar:/slf4j-api-1.7.25.jar:/slf4j-simple-1.7.25.jar org.codenarc.CodeNarc -report=html -report=xml -basedir=jenkins-pipelines'
-                } finally {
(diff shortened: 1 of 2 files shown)


Jenkinsfile.codenarc
```diff
@@ -10,35 +10,17 @@ githubCollaboratorCheck(
...
-            container('codenarc') {
-                try {
-                    sh 'java -classpath /groovy-all-2.4.6.jar:/CodeNarc-1.1.jar:/slf4j-api-1.7.25.jar:/slf4j-simple-1.7.25.jar org.codenarc.CodeNarc -report=html -report=xml -basedir=jenkins-pipelines'
-                } finally {
-                    archiveArtifacts(artifacts: "CodeNarcReport.html", fingerprint: true)
-                    warningsCodeNarc(filename: 'CodeNarcXmlReport.xml')
-                }
+    stage('Style Checks') {
+        try {
+            sh("docker run -v ${WORKSPACE}:/codenarc --tty --rm registry.suse.de/devel/casp/ci/opensuse_leap_42.3_containers/jenkins-codenarc-container:latest sh -c 'cd /codenarc; java -classpath /groovy-all-2.4.6.jar:/CodeNarc-1.1.jar:/slf4j-api-1.7.25.jar:/slf4j-simple-1.7.25.jar org.codenarc.CodeNarc -report=html -report=xml -basedir=jenkins-pipelines'")
```
ereslibre: Usually jenkins doesn't like the `--tty` option because the worker does not allocate one.

Can't we use the `docker` `agent` directive instead of sh to `docker` here?

hwoarang: actually scratch that. The docker pipeline is a hard dep of the pipeline plugin so we can't avoid it.

hwoarang: @ereslibre better now?

Jenkinsfile.codenarc
```diff
@@ -10,34 +10,18 @@ githubCollaboratorCheck(
...
-                envVar(key: 'http_proxy', value: env.http_proxy),
-                envVar(key: 'https_proxy', value: env.http_proxy),
-            ],
-        ),
-]) {
-    node(label) {
-        stage('Retrieve Code') {
-            checkout scm
-        }
+node("leap15.0" && "caasp-pr-worker") {
```
MalloZup: Nit we could put node in a var,

Jenkinsfile.codenarc
```diff
@@ -10,34 +10,18 @@ githubCollaboratorCheck(
...
         stage('Style Checks') {
-            container('codenarc') {
-                try {
-                    sh 'java -classpath /groovy-all-2.4.6.jar:/CodeNarc-1.1.jar:/slf4j-api-1.7.25.jar:/slf4j-simple-1.7.25.jar org.codenarc.CodeNarc -report=html -report=xml -basedir=jenkins-pipelines'
-                } finally {
-                    archiveArtifacts(artifacts: "CodeNarcReport.html", fingerprint: true)
-                    warningsCodeNarc(filename: 'CodeNarcXmlReport.xml')
-                }
+            try {
+                sh 'java -classpath /groovy-all-2.4.6.jar:/CodeNarc-1.1.jar:/slf4j-api-1.7.25.jar:/slf4j-simple-1.7.25.jar org.codenarc.CodeNarc -report=html -report=xml -basedir=/codenarc/jenkins-pipelines'
```
MaximilianMeister: now that we're on it, maybe we could make those invocations more consistent to be in line with the other `sh(script: "")` calls?

PR #610, opened 2019-02-05
[3.0] velum-bootstrap: check content disposition instead of url

vitoravelino:
The spec for checking the kubeconfig download was previously done
through checking an url associated with the automatic download.

A better way of checking the automatic download is to check the
content disposition in the response header.

The previous behavior to manually download kubeconfig was maintained in
order to use the kubeconfig in further tests. The only difference is
that is compatible with both GET and POST requests due to different
versions in different repos. In the future we can remove the GET and
keep only the POST one.

(cherry picked from commit 3a1afa4127ff2caa303057817e63ab3a92aa7a90)

Backport of #592


vitoravelino:
Forgot to backport this in order to make https://github.com/SUSE/kubic-velum-security-fixes/pull/9 green.

vitoravelino:
Can we merge this?

PR #617, opened 2019-02-08
misc-tools: upload-image-openstack: Fix conditional for glance images

hwoarang:
The openstack command always returns 0 if the command was successful
even when an image was not present. As such, we need to actually check
the output of this command to find out if we already have the image on
the server.


hwoarang:
Self merging because it does not affect regular PR tests.

PR #633, opened 2019-02-12
Heat: Add some outputs to master/worker templates

rhafer:
By having these outputs available we can simplify some tasks in the
SOC on Kubernetes (socok8s) project (e.g. populating the inventory).


hwoarang:
@mjura @tdaines42 could you have a look at this please?

PR #636, opened 2019-02-12
rename FILE variable to ENVIRONMENT_JSON in testAllRoles.sh

lcavajani:

## What does this PR change?
- How does this PR fixes it?

ENVIRONMENT_JSON was not used before, this is just a bit cleaner.


lcavajani:
done

PR #639, opened 2019-02-13
removed ses hardcoded repos, added ses reg code

ellisab:
## What does this PR change?
It was using ses hardcoded repos
- How does this PR fixes it?
Now it uses ses registration code

### DIFF
--- a/ses-openstack-terraform/deepsea.sh
+++ b/ses-openstack-terraform/deepsea.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+sleep 1m
 zypper -n in -l -y ntp 
 
 cat <<EOF > /etc/ntp.conf
--- a/ses-openstack-terraform/deepsea.sh
+++ b/ses-openstack-terraform/deepsea.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-sleep 1m
+while [ "$(zypper refresh > /dev/null 2>&1; echo $?)" -ne 0 ]; do 
+    printf "wait 5 seconds for admin to be registered to scc\n"
+    sleep 5
+done
+
 zypper -n in -l -y ntp 
 
 cat <<EOF > /etc/ntp.conf
--- a/ses-openstack-terraform/mon.tpl
+++ b/ses-openstack-terraform/mon.tpl
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-SUSEConnect -r ${regcode}
-zypper --non-interactive --quiet --gpg-auto-import-keys ar -f ${ses_base} ses_base
-zypper --non-interactive --quiet --gpg-auto-import-keys ar -f ${ses_update} ses_update
+SUSEConnect -r ${regcodeSLES}
+sleep 1
+SUSEConnect -p ses/5/x86_64 -r ${regcodeSES}
+sleep 1
 
 zypper -n in -l -y ntp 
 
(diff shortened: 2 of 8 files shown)


ses-openstack-terraform/deepsea.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+sleep 1m
```
hwoarang: what is this :) ? Is it unrelated?

ses-openstack-terraform/mon.tpl
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-SUSEConnect -r ${regcode}
-zypper --non-interactive --quiet --gpg-auto-import-keys ar -f ${ses_base} ses_base
-zypper --non-interactive --quiet --gpg-auto-import-keys ar -f ${ses_update} ses_update
+SUSEConnect -r ${regcodeSLES}
+sleep 1
```
hwoarang: Why are these sleep statements necessary? Is there no better way to do this? Like a while look to check whether the product has been registered?

ses-openstack-terraform/ses-cluster.tf
```diff
@@ -365,7 +362,7 @@ resource "openstack_compute_floatingip_associate_v2" "mon_ext_ip" {
 
 resource "openstack_blockstorage_volume_v2" "osd-blk" {
   count = "${var.osds}"
-  size  = 40
+  size  = "${var.volumeSize}"
```
hwoarang: So this is a nice change but it is unrelated. Please move it to a separate commit

ses-openstack-terraform/ses-openstack
```diff
@@ -87,4 +87,4 @@ fi
 
 terraform $COMMAND -var auth_url=$OS_AUTH_URL -var domain_name=$OS_USER_DOMAIN_NAME \
                     -var region_name=$OS_REGION_NAME -var project_name=$OS_PROJECT_NAME \
-                    -var user_name=$OS_USERNAME -var password=$OS_PASSWORD -var-file=openstack.tfvars
+                    -var user_name=$OS_USERNAME -var password=$OS_PASSWORD -var-file=terraform.tfvars
```
hwoarang: Similarly this also feels like an unrelated change to please move it to a separate commit

ses-openstack-terraform/deepsea.sh
ellisab: It avoids zypper locked error

ses-openstack-terraform/mon.tpl
ellisab: It avoids a zypper locked error

ses-openstack-terraform/deepsea.sh
ereslibre: Can we try to do an active wait here for something that is more descriptive? I wonder how it avoids the zypper lock error.  This itself can be a problem (besides the random wait), we can hit a zypper lock error again even after passing this, right? If something else locked zypper? Maybe one option would be to even loop outside the `zypper -n in` command until it succeeds (with a limit of retries)

ses-openstack-terraform/mon.tpl
ereslibre: Same comment as before, I agree with @hwoarang, and I'd prefer to do a more explicit conditional on what we are waiting for, or even retrying what we expect to succeed.

ses-openstack-terraform/deepsea.sh
drpaneas: In openQA we used to disable the packagekit as it can start out of the blue, blocking with zypper. The command we used to do that is:

```
systemctl mask packagekit; systemctl stop packagekit; while pgrep packagekitd; do sleep 1; done
```

Dunno if it helps, I just wanted to inform you about it as well :)

PR #640, opened 2019-02-14
misc-tools: upload-image-openstack: Fix wrong variable name

hwoarang:
The correct variable for the iteration is 'image' so fix it. This fixes
the following problem in the CI

[...]
[+] Deleting previous SUSE CaaSP qcow2 VM image for {version=4.0,channel=devel}
./upload-image-openstack.sh: line 25: images: unbound variable


hwoarang:
Self merging because this does not affect any regular job really

PR #642, opened 2019-02-14
WIP: add snapshot tool for devenv

MalloZup:
# What change this pr?

Main motivation of this `util` is to have a snapshot util for speeding up local dev env.

so an user could do after deployment:

`./misc-tools/snapshot-devenv.sh` ( creating snapshots for vms)

`./misc-tools/snapshot-devenv.sh revert` this will revert vms to snapshots

`./misc-tools/snapshot-devenv.sh delete` this remove snapshots.

# todo:

- [ ]  remove hardcoded list and use a `jq` call


MalloZup:
yep i can integrate it to caasp-devenv actually

misc-tools/snapshot-devenv.sh
```diff
@@ -0,0 +1,112 @@
...
+action=$1
+
+# default create snapshot
+if [ -z "$action" ]; then
+   usage
+fi
+
+if [[ $action = "create" ]]
+then
+    create-snapshots $caasp_vms
```
dannysauer: Does this actually work?  I'd expect it to only return the first element of the array.

I'm prety sure that, in bash, to pass an array in to a function, you have two options.  One is to pass in all of the elements:
```bash
function process_array {
  for element in "$[@]"
  do
}
process_array "${some_array[@]}"
```

Or, pass the array in using a nameref - which is as close to a pointer as Bash gets:
```bash
function process_arrayref {
  declare -n arr=$1
  for element in "${n[@]}"
  do...
}
process_arrayref "name_of_array"
```

misc-tools/snapshot-devenv.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# get from CLI the action
+action=$1
+
+# default create snapshot
+if [ -z "$action" ]; then
+   usage
+fi
+
+if [[ $action = "create" ]]
```
dannysauer: This (and the previous if) could all be a case statement. :)

```bash
case "${action:-}" in
  "create" ) create-snapshots args ;;
  "delete" ) delete-snapshots args ;;
  "revert" ) revert-snapshots args ;;
  *)         usage ;;
esac

File: misc-tools/snapshot-devenv.sh (hunk 1 above)
Comment (MalloZup) [reply misc-tools/snapshot-devenv.sh]: i have tested locally worked for me. i  will consider your suggestion. thx :+1:

File: misc-tools/snapshot-devenv.sh (hunk 2 above)
Comment (MalloZup) [reply misc-tools/snapshot-devenv.sh]: i think i will remove this because the plan imho is integrate to main `cli`.

so basically `caasp-devenv --snapshot delete` or `caasp-devenv --snapshot create`

thx for review, anyways :rocket:

File: misc-tools/snapshot-devenv.sh [hunk 3]
```diff
@@ -0,0 +1,112 @@
...
+ create 1 snapshots of all existing caasp vms
+
+02 ./snapshot-devenv  revert
+ revert your existings caasp vms to the snapshot created
+
+03 ./snapshot-devenv  delete
+ delete caasp vms snapshots
+USAGE
+)
+echo "$USAGE"
```
dannysauer: Wouldn't it be easier to just do the cat directly instead of capturing cat's stdout in a subshell to populate a variable and which is immediately printed to stdout? ;)

misc-tools/snapshot-devenv.sh
dannysauer: (sorry about the updates; I wasn't paying close enough attention when I typed the first time)

misc-tools/snapshot-devenv.sh
dannysauer: Yeah, I like the idea of integrating into the main code.  I figured this might still be useful there, though. :D

misc-tools/snapshot-devenv.sh
MalloZup: thhx fore review appreiciated

misc-tools/snapshot-devenv.sh
MalloZup: donee thx

misc-tools/snapshot-devenv.sh
MalloZup: fixed

PR #646, opened 2019-02-15
remove obsolete openstack.tfvars, set sensible value for volumeSize, …

ellisab:
…add terraform -auto-approve flag, document regcodeSES variable

## What does this PR change?
openstack.tfvars is no longer used, tf -auto-approve skips yes step, volumeSize should have a sensible value for installing CAP
- How does this PR fixes it?
remove obsolete openstack.tfvars, set sensible value for volumeSize, add terraform -auto-approve flag, document regcodeSES variable


ses-openstack-terraform/terraform.tfvars
```diff
@@ -9,4 +9,4 @@ osds = 5
 regcodeSLES = "sles"
 regcodeSES = "ses"
 identifier = "test"
-volumeSize = 1
\ No newline at end of file
+volumeSize = 25
```
lcavajani: we should keep the variables formatted in snake case

ses-openstack-terraform/terraform.tfvars
```diff
@@ -9,4 +9,4 @@ osds = 5
 regcodeSLES = "sles"
```
lcavajani: we should keep the variables formatted in snake case

ses-openstack-terraform/terraform.tfvars
```diff
@@ -9,4 +9,4 @@ osds = 5
...
 regcodeSES = "ses"
```
lcavajani: same here

ses-openstack-terraform/terraform.tfvars
ellisab: volumeSize should be volume_size then? What do the others think about this?

ses-openstack-terraform/terraform.tfvars
ellisab: recodeSLES should be regcode_sles or regcode_SLES? What do the others think about this?

ses-openstack-terraform/terraform.tfvars
MalloZup: i think snake case make sense if the whole logic use snake case already

ses-openstack-terraform/terraform.tfvars
```diff
@@ -6,7 +6,7 @@ master_size = "m1.large"
...
 worker_size = "m1.large"
 osds = 5
-regcodeSLES = "sles"
-regcodeSES = "ses"
+regcode_SLES = "sles"
+regcode_SES = "ses"
 identifier = "test"
-volumeSize = 1
\ No newline at end of file
+volumeSize = 25
```
MalloZup: we can do a separate pr for renaming to `volume_size`

lcavajani: what's the point of doing it in an other PR ? I think it is the same scope

ses-openstack-terraform/terraform.tfvars
```diff
@@ -6,7 +6,7 @@ master_size = "m1.large"
 mons = 3
 worker_size = "m1.large"
 osds = 5
-regcodeSLES = "sles"
-regcodeSES = "ses"
+regcode_SLES = "sles"
```
lcavajani: I am being picky here but I am in favor of keeping everything in lower case so `regcode_sles`, `regcode_ses` so we don't have mix cased variables

ses-openstack-terraform/terraform.tfvars
MalloZup: i'm ok with both

ses-openstack-terraform/terraform.tfvars
ellisab: given that our products abbreviations are in upper case i favor upper case

lcavajani: this is not about how are our products are named in real life, in documentation of course I would use `SLES` and `SES`, it is about keeping keeping the same variable formatting

